PR #31, opened 2019-06-26
Added --runInBand for Jest on Circle

JoshuaKGoldberg:
https://jestjs.io/docs/en/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server


JoshuaKGoldberg:
Heh, didn't really change timing at all.

PR #67, opened 2019-07-02
Replace fs promises api with utils.promisify

### LINKED ISSUE #63: Support Node 8.X
### 🚀 Feature Request

Node 8 is still in LTS through the end of 2019 _(though Node 6 is not anymore, hooray!)_. `tslint-eslint-config` should explicitly support it.

#### Existing Behavior

Instead, `tslint-eslint-config` crashes silently when run in Node 8. Darn.

#### Change Proposal

Usage of the new `.promises` APIs, e.g. `fs.promises` (https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fs_promises_api), is *experimental* in Node 10 and not available in Node 8. Any usage of this:

```ts
import * as fs from "fs";

// ...

await fs.promises.readFile("...");
```

...should be replaced by something like:

```ts
import * as fs from "fs";
import { promisify } from "utils";

const readFile = promisify(fs.readFile);

// ...

await readFile("...");
```

egiurleo:
## PR Checklist

-   [x] Addresses an existing issue: fixes #63
-   [x] That issue was marked as [`status: accepting prs`](https://github.com/JoshuaKGoldberg/tslint-to-eslint-config/labels/status%3A%20accepting%20prs)

## Overview
This PR replaces the `fs.promises` API with `util.promisify` in order to support Node 8.x.

Edit: Not sure how to test this change.

### DIFF
--- a/src/adapters/fsFileSystem.ts
+++ b/src/adapters/fsFileSystem.ts
@@ -4,6 +4,7 @@ import { promisify } from "util";
 import { FileSystem } from "./fileSystem";
 
 const readFile = promisify(fs.readFile);
+const writeFile = promisify(fs.writeFile);
 
 export const fsFileSystem: FileSystem = {
     fileExists: async (filePath: string) => {
@@ -22,7 +23,7 @@ export const fsFileSystem: FileSystem = {
     },
     writeFile: async (filePath: string, contents: string) => {
         try {
-            return fs.promises.writeFile(filePath, contents);
+            return writeFile(filePath, contents);
         } catch (error) {
             return error;
         }
--- a/src/adapters/fsFileSystem.ts
+++ b/src/adapters/fsFileSystem.ts
@@ -3,6 +3,8 @@ import { promisify } from "util";
 
 import { FileSystem } from "./fileSystem";
 
+const readFile = promisify(fs.readFile);
+
 export const fsFileSystem: FileSystem = {
     fileExists: async (filePath: string) => {
         try {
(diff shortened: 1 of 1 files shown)


src/adapters/fsFileSystem.ts
```diff
@@ -12,7 +13,8 @@ export const fsFileSystem: FileSystem = {
     },
     readFile: async (filePath: string) => {
         try {
-            return (await fs.promises.readFile(filePath)).toString();
+            const readFile = promisify(fs.readFile);
```
JoshuaKGoldberg: Nit: would you mind moving this variable to the top of the file? That way we don't `promisify` each time this gets called.

JoshuaKGoldberg:
For testing this change: to recap an offline conversation, exercising that file writing works could be done with a single `tslint-to-eslint-config` run that writes a new config file.

src/adapters/fsFileSystem.ts
```diff
@@ -12,14 +16,14 @@ export const fsFileSystem: FileSystem = {
...
-            return (await fs.promises.readFile(filePath)).toString();
+            return (await readFile(filePath)).toString();
         } catch (error) {
             return error;
         }
     },
     writeFile: async (filePath: string, contents: string) => {
         try {
-            return fs.promises.writeFile(filePath, contents);
+            return writeFile(filePath, contents);
```
egiurleo: @JoshuaKGoldberg let me know if this looks right! (write ;D)

JoshuaKGoldberg: @egiurleo perfect 👌🕺💃

PR #87, opened 2019-07-06
Print Rule Notices

### LINKED ISSUE #7: Print rule notices in result reporting
Rule converters may contain a `notices` member intended to message users about behavior changes from TSLint to ESLint. For example, `no-arg`'s converter returns a notice about `arguments.callee` also being banned.

Unfortunately, there's nothing _(yet!)_ that prints these notices in the CLI. `reportConversionResults` should have code added to also print out, for each resultant ESLint rule with notices, the notices under it.

Maybe the output could look something like this?

```
📢 2 ESLint rules behave differently from their TSLint counterparts: 📢
    * no-caller:
        - `arguments.callee` will now also be banned.
    * no-invalid-this:
        - Functions in methods will no longer be ignored.
```

egiurleo:
## PR Checklist

-   [x] Addresses an existing issue: fixes #7 
-   [x] That issue was marked as [`status: accepting prs`](https://github.com/JoshuaKGoldberg/tslint-to-eslint-config/labels/status%3A%20accepting%20prs)

## Overview
This PR updates `reportConversionResults` to print out rule notices.


src/reporting/reportConversionResults.ts
```diff
@@ -93,3 +94,28 @@ const logMissingPlugins = (plugins: Set<string>, logger: Logger) => {
...
+    if (rulesWithNotices.length > 0) {
+        logger.stdout.write(chalk.yellowBright(`📢 ${rulesWithNotices.length} ESLint`));
+        logger.stdout.write(
+            chalk.yellowBright(rulesWithNotices.length == 1 ? ` rule behaves` : ` rules behave`),
+        );
+        logger.stdout.write(
+            chalk.yellowBright(` differently from their TSLint counterparts: 📢${EOL}`),
+        );
+
+        rulesWithNotices.forEach(rule => {
```
egiurleo: There might be a more idiomatic way to write this nested for loop :)

src/rules/convertRules.ts
```diff
@@ -75,6 +75,7 @@ export const convertRules = (
                         existingConversion.ruleArguments,
                         newConversion.ruleArguments,
                     ),
+                    notices: merger(existingConversion.notices, newConversion.notices),
```
egiurleo: I'm not sure if there's ever a case where two rules will have different notices that need to be merged? There's a unit test failing on this branch that's related to this. Let's talk more about it and then I'll fix the test :)

src/reporting/reportConversionResults.ts
JoshuaKGoldberg: Yeah I would move the `logger.stdout.write` to inside the `if (rule.notices !== undefined) {` block. There's not really a need to print out that line.

However, if you change `ruleOptions =>` to something like `(ruleOptions): ruleOptions is RuleWithNotices`, where that's a type that says `.notices` isn't `undefined`, you can skip the `if` statement altogether. Let's go with that to simplify the output JS code. https://fettblog.eu/typescript-type-predicates

src/reporting/reportConversionResults.ts
```diff
@@ -93,3 +94,28 @@ const logMissingPlugins = (plugins: Set<string>, logger: Logger) => {
...
+            logger.stdout.write(chalk.yellow(`* ${rule.ruleName}:${EOL}`));
+            if (rule.notices !== undefined) {
+                rule.notices.forEach(notice => {
+                    logger.stdout.write(chalk.yellow(`- ${notice}${EOL}`));
```
JoshuaKGoldberg: It'd be a little difficult to tell visual hierarchies in the output when the `*` has the same indentation as the `-`.
```suggestion
                    logger.stdout.write(chalk.yellow(`    - ${notice}${EOL}`));
```

src/rules/convertRules.ts
JoshuaKGoldberg: Oh that's a really good point I didn't think of. For now, can you just do `[...existingConversion.notices, ...newConversion.notices]`, and file a followup issue to investigate? I suspect we'll want to expand mergers into `merger.ruleArguments` and `merger.notices`, but agreed: +1 to talking more about it.

In the spirit of being agile, IMO we should get your changes in soon and do that as a followup.

src/reporting/reportConversionResults.ts
egiurleo: Nice! Thanks!

egiurleo:
@JoshuaKGoldberg made the changes you requested! Check it out!

src/reporting/reportConversionResults.ts
```diff
@@ -93,3 +94,31 @@ const logMissingPlugins = (plugins: Set<string>, logger: Logger) => {
...
+
+interface RuleWithNotices {
+    notices: any[];
+    ruleName: string;
+}
+
+const logNotices = (converted: Map<string, ESLintRuleOptions>, logger: Logger) => {
+    const rulesWithNotices = Array.from(converted.values()).filter(
+        ruleOptions => ruleOptions.notices && ruleOptions.notices.length >= 1,
+    ) as RuleWithNotices[];
```
egiurleo: @JoshuaKGoldberg is this kind of what you had in mind? (Essentially casting each rule within the new array)

JoshuaKGoldberg: Yes, pretty much 🚀 thanks!

JoshuaKGoldberg: As a post-merge FYI - the way I was thinking of was:

```ts
const rulesWithNotices = Array.from(converted.values()).filter(
    (ruleOptions): ruleOptions is RuleWithNotices =>
        !!(ruleOptions.notices && ruleOptions.notices.length >= 1)
);
```

...which IMO is _slightly_ preferable because using `as` casts is kind of a way around the type system (rather than informing it of the stuff we already know). But this is such a small difference that it doesn't seem reasonable to block the PR on it.

PR #137, opened 2019-08-06
converted explicit-member-accessibility will ignore constructors

eranshabi:
## Overview
when converting `member-access` to `@typescript-eslint/explicit-member-accessibility`, it should configure it to ignore constructors like it does for the tslint rule


JoshuaKGoldberg:
Hi @eranshabi, thanks for sending this PR in! 🙌 

PR builds aren't running right now (#138), but I've verified locally the linter, compiler, and tests pass. Looks great!

Filed #139 to track the rest of the rule's options as well.

JoshuaKGoldberg:
Published in `0.2.3`! 🎈🎊

PR #141, opened 2019-08-12
Merge and deduplicate rule notices

### LINKED ISSUE #119: Investigate merging notices
In #87, `tslint-to-eslint-config` began logging rule notices, but there wasn't any infrastructure that took into account the possibility of having to merge notices. I have updated line 81 in `convertRules.ts` to just merge the two notice arrays, but we should investigate creating a merger for rule notices.

egiurleo:
Fixes #119

### DIFF
--- a/src/rules/convertRules.test.ts
+++ b/src/rules/convertRules.test.ts
@@ -95,7 +95,6 @@ describe("convertRules", () => {
                     {
                         ruleName: "eslint-rule-a",
                         ruleSeverity: "error",
-                        notices: [],
                     },
                 ],
             ]),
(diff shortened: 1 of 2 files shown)


src/rules/convertRules.ts
```diff
@@ -69,16 +73,15 @@ export const convertRules = (
...
-
                 converted.set(changes.ruleName, {
                     ...existingConversion,
                     ruleArguments: merger(
                         existingConversion.ruleArguments,
                         newConversion.ruleArguments,
                     ),
-                    notices: [...existingNotices, ...newNotices],
+                    notices: Array.from(
+                        new Set([...existingConversion.notices, ...newConversion.notices]),
```
egiurleo: Let me know if this is better than it was previously (setting a default value for notices earlier on vs in the `existingNotices` and `newNotices` variables.

src/rules/convertRules.test.ts
```diff
@@ -183,6 +184,51 @@ describe("convertRules", () => {
         );
     });
 
+    it("merges and deduplicates rule notices", () => {
```
egiurleo: This test is mostly copy pasta. I don't think it has anything unnecessary but happy to remove stuff if I missed something :)

egiurleo:
@JoshuaKGoldberg take a look :)

src/rules/convertRules.ts
JoshuaKGoldberg: @egiurleo unfortunately no, if `newNotices` is `undefined`, this will throw an error:

Trying out in a Node CLI:

```ts
> [...undefined, ...[]]
Thrown:
TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
```

...which seems like a missing unit test!

src/rules/convertRules.test.ts
JoshuaKGoldberg: Yeah, I think this file has a lot of [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)/[WET](https://dev.to/wuz/stop-trying-to-be-so-dry-instead-write-everything-twice-wet-5g33) violations. Either the source function is doing too much and/or the unit tests should have common logic moved into helpers.

A few options for you here, all of which I have no issue with:
* Ignore this comment and continue eating the 🍝
* File a followup issue to investigate this
* In this PR, move shared logic from these unit tests into a function in this file
* In this PR, refactor some logic out of `convertRules` so edge cases can be directly unit tested - for example, the contents of the `for` loop.

egiurleo: Gonna ignore for now and create a follow-up issue.

src/rules/convertRules.test.ts
```diff
@@ -183,6 +183,96 @@ describe("convertRules", () => {
...
+        };
+        const conversionResult = {
+            rules: [
+                {
+                    ruleName: "eslint-rule-a",
+                    notices: undefined,
+                },
+                {
+                    ruleName: "eslint-rule-a",
+                    notices: undefined,
```
egiurleo: @JoshuaKGoldberg is this the test case you're looking for? (Will make a follow-up PR to DRY out these tests)

JoshuaKGoldberg: Yeah, this is perfect! 🙌

JoshuaKGoldberg:
Released in `0.2.4`! ✨

PR #192, opened 2019-10-02
Added missing quotemark converter

### LINKED ISSUE #182: Missing converter: quotemark
### 💥 Missing Converter

#### TSLint Rule Name

`quotemark`

mauricetmeyer:
## PR Checklist

-   [x] Addresses an existing issue: fixes #182

## Overview

Adds a missing rule converter for the `quotemark` rule.

### DIFF
--- a/src/rules/converters/quotemark.ts
+++ b/src/rules/converters/quotemark.ts
@@ -1,0 +1,32 @@
+import { RuleConverter } from "../converter";
+
+export const convertQuotemark: RuleConverter = tslintRule => {
+    const notices: string[] = [];
+    const ruleArguments: any[] = [];
+
+    ["jsx-single", "jsx-double", "avoid-template"].forEach(rule => {
+        if (tslintRule.ruleArguments.includes(rule)) {
+            notices.push(`Option "${rule}" is not supported by ESLint.`);
+        }
+    });
+
+    ["single", "double", "backtick"].forEach(rule => {
+        if (tslintRule.ruleArguments.includes(rule)) {
+            ruleArguments.push(rule);
+        }
+    });
+
+    if (tslintRule.ruleArguments.includes("avoid-escape")) {
+        ruleArguments.push({ avoidEscape: true });
+    }
+
+    return {
+        rules: [
+            {
+                notices,
+                ruleArguments,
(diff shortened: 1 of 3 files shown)


src/rules/converters/quotemark.ts
```diff
@@ -0,0 +1,32 @@
+import { RuleConverter } from "../converter";
+
+export const convertQuotemark: RuleConverter = tslintRule => {
+    const notices: string[] = [];
+    const ruleArguments: any[] = [];
```
JoshuaKGoldberg: `any` is kind of a dangerous bad kid in TypeScript land. It removes a lot of type safety from its operations. Let's switch to the more specific `string[]`:

```suggestion
    const ruleArguments: string[] = [];
```

src/rules/converters/quotemark.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    ["jsx-single", "jsx-double", "avoid-template"].forEach(rule => {
```
JoshuaKGoldberg: > `rule`

This confused me for a few seconds: really we're looping over the _options_ for the rule, not rules themselves. Could you please rename these `rule` variables to `option`?

src/rules/converters/tests/quotemark.test.ts
```diff
@@ -0,0 +1,87 @@
...
+                {
+                    ruleName: "@typescript-eslint/quotes",
+                    ruleArguments: [],
+                    notices: [],
+                },
+            ],
+        });
+    });
+
+    test("conversion with an arguments", () => {
```
JoshuaKGoldberg: 😉 

```suggestion
    test("conversion with arguments", () => {
```

src/rules/converters/quotemark.ts
mauricetmeyer: Done

src/rules/converters/quotemark.ts
mauricetmeyer: I can't use string[] as avoidEscape rule is an object, but I created the type below that is now used as the array type.

```ts
type QuotemarkRule = string | { avoidEscape: true }
const ruleArguments: QuotemarkRule[] = [];
```

PR #193, opened 2019-10-02
add converter for space-within-parens rule

### LINKED ISSUE #184: Missing converter: space-within-parens
### 💥 Missing Converter

#### TSLint Rule Name

`space-within-parens`

bmeverett:
## PR Checklist

-   [x] Addresses an existing issue: fixes #184 


src/rules/converters/space-within-parens.ts
```diff
@@ -0,0 +1,14 @@
+import { RuleConverter } from "../converter";
+
+export const convertSpaceWithinParens: RuleConverter = tslintRule => {
+    return {
+        rules: [
+            {
+                ...(tslintRule.ruleArguments.length !== 0 && {
+                    ruleArguments: tslintRule.ruleArguments,
```
JoshuaKGoldberg: https://eslint.org/docs/rules/space-in-parens doesn't show the ESLint `space-in-parens` rule as being able to receive a numeric literal the same way as the TSLint rule. It instead has `always` or `never` as options. This'll need to convert from the TSLint way to the ESLint way.

bmeverett: Ah, I was looking at the tslint docs and didn't look at eslint. I will update this to convert appropriately.

src/rules/converters.ts
```diff
@@ -222,12 +224,12 @@ export const converters = new Map([
...
-    // Some of them will likely need to have notices about changed lint behaviors...
-    // If you're willing to take on that work, that'd be great! Please send PRs! 💖
-    // As these are enabled, they should be added in sorted order to the list above.
+    // these converters are all for rules that need more complex option conversions.
+    // some of them will likely need to have notices about changed lint behaviors...
+    // if you're willing to take on that work, that'd be great! Please send PRs! 💖
+    // as these are enabled, they should be added in sorted order to the list above.
 
-    // TSLint core rules:
+    // tSLint core rules:
```
JoshuaKGoldberg: Looks like these comments were auto-formatted to lowercase? Let's revert them.

src/rules/converters/space-within-parens.ts
```diff
@@ -0,0 +1,18 @@
+import { RuleConverter } from "../converter";
+
+export const convertSpaceWithinParens: RuleConverter = tslintRule => {
+    let arg = "";
+    if (tslintRule.ruleArguments.length === 1) {
+        arg = "always";
+    } else {
+        arg = "never";
+    }
```
JoshuaKGoldberg: Or more simply:

```ts
const arg = tslintRule.ruleArguments.length === 1 ? "always" : "never";
```

...as the temporary `""` is never used and a bit misleading.

src/rules/converters/space-within-parens.ts
```diff
@@ -0,0 +1,18 @@
+import { RuleConverter } from "../converter";
+
+export const convertSpaceWithinParens: RuleConverter = tslintRule => {
+    let arg = "";
+    if (tslintRule.ruleArguments.length === 1) {
+        arg = "always";
```
JoshuaKGoldberg: This is actually a behavior change from TSLint to ESLint. TSLint allows specifying a specific number of parenthesis, while ESLint enforces one (at least I think it's exactly one, not at least one - it's a little hard to tell from the docs). Could you please add a `notices` explaining that?

src/rules/converters.ts
```diff
@@ -222,10 +224,10 @@ export const converters = new Map([
...
     ["triple-equals", convertTripleEquals],
 
-    // These converters are all for rules that need more complex option conversions.
-    // Some of them will likely need to have notices about changed lint behaviors...
-    // If you're willing to take on that work, that'd be great! Please send PRs! 💖
-    // As these are enabled, they should be added in sorted order to the list above.
+    // these converters are all for rules that need more complex option conversions.
+    // some of them will likely need to have notices about changed lint behaviors...
+    // if you're willing to take on that work, that'd be great! Please send PRs! 💖
+    // as these are enabled, they should be added in sorted order to the list above.
```
JoshuaKGoldberg: Mind reverting these changes?

src/rules/converters.ts
```diff
@@ -236,7 +238,8 @@ export const converters = new Map([
     // ["no-trailing-whitespace", convertNoTrailingWhitespace], //  no-trailing-spaces
     // ["no-unused-expression", convertNoUnusedExpression], // no-unused-expressions
     // ["no-void-expression", convertNoVoidExpression], // (no exact equivalent)
-    // ["space-within-parens", convertSpaceWithinParens], // space-in-parens
+    // ["quotemark", convertQuotemark], // quotes
+    // ["triple-equals", convertTripleEquals], // eqeqeq
```
JoshuaKGoldberg: Looks like add an odd merge conflict? 😊

src/rules/converters/space-within-parens.ts
```diff
@@ -0,0 +1,14 @@
+import { RuleConverter } from "../converter";
+
+export const convertSpaceWithinParens: RuleConverter = tslintRule => {
+    const arg = tslintRule.ruleArguments.length === 1 ? "always" : "never";
+    return {
+        rules: [
+            {
+                ruleArguments: [arg],
+                ruleName: "@typescript-eslint/space-within-parens",
+                notices: ["The number of spaces will be ignored"],
```
JoshuaKGoldberg: Ah, almost - this notice isn't relevant if the user has configured to `never` allow spaces, right? As in, the rule behavior then is the same in the two linters. It's just when there are spaces in play that they differ?

bmeverett: So should the notice only show when parameters are passed then?

JoshuaKGoldberg: I believe so, yeah.

PR #194, opened 2019-10-03
Adds no-for-in converter along with tests

### LINKED ISSUE #170: Missing converter: no-for-in
### 💥 Missing Converter

#### TSLint Rule Name

`no-for-in`

bpallares:
## PR Checklist

-   [x] Addresses an existing issue: fixes #170

## Overview

 Adds no-for-in converter

### DIFF
--- a/src/rules/converters/no-for-in.ts
+++ b/src/rules/converters/no-for-in.ts
@@ -1,0 +1,14 @@
+import { RuleConverter } from "../converter";
+
+export const convertNoForIn: RuleConverter = tslintRule => {
+    return {
+        rules: [
+            {
+                ...(tslintRule.ruleArguments.length !== 0 && {
+                    ruleArguments: tslintRule.ruleArguments,
+                }),
+                ruleName: "no-restricted-syntax",
+            },
+        ],
+    };
+};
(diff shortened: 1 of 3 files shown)


src/rules/converters/no-for-in.ts
```diff
@@ -0,0 +1,14 @@
+import { RuleConverter } from "../converter";
+
+export const convertNoForIn: RuleConverter = tslintRule => {
+    return {
+        rules: [
+            {
+                ...(tslintRule.ruleArguments.length !== 0 && {
+                    ruleArguments: tslintRule.ruleArguments,
+                }),
```
JoshuaKGoldberg: Hmm, this isn't quite right. [`no-restricted-syntax`](https://eslint.org/docs/rules/no-restricted-syntax) takes in a list of things it's meant to ban, which are each either a string selector or an object with a string selector and string message. `no-for-in` doesn't provide anything like that.

You'll want to provide a fairly `ruleArguments` that just says to avoid the `for-in` syntax. It looks like your tests start to do that?

bpallares: Oh , so basically we will be good if we fix the `ruleArguments` to `["error", "ForInStatement"]` ?

JoshuaKGoldberg: Yup!

PR #202, opened 2019-10-04
Add missing converter: no-null-keyword

### LINKED ISSUE #173: Missing converter: no-null-keyword
### 💥 Missing Converter

#### TSLint Rule Name

`no-null-keyword`

da-snap:
Add converter for #173

## PR Checklist

-   [x] Addresses an existing issue: fixes #173 

## Overview

Add converter to convert `no-null-keyword` rule. Will add plugin eslint-plugin-no-null to settings.

### DIFF
--- a/src/rules/converters/no-null-keyword.ts
+++ b/src/rules/converters/no-null-keyword.ts
@@ -2,6 +2,7 @@ import { RuleConverter } from "../converter";
 
 export const convertNoNullKeyword: RuleConverter = () => {
     return {
+        notices: ["Null types will no longer be handled."],
         rules: [
             {
                 ruleName: "no-null/no-null",
--- a/src/rules/converters/no-null-keyword.ts
+++ b/src/rules/converters/no-null-keyword.ts
@@ -1,0 +1,12 @@
+import { RuleConverter } from "../converter";
+
+export const convertNoNullKeyword: RuleConverter = () => {
+    return {
+        rules: [
+            {
+                ruleName: "no-null/no-null",
+            },
+        ],
+        plugins: ["no-null"],
+    };
+};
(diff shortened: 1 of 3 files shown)


src/rules/converters/no-null-keyword.ts
```diff
@@ -0,0 +1,12 @@
+import { RuleConverter } from "../converter";
+
+export const convertNoNullKeyword: RuleConverter = () => {
+    return {
+        rules: [
+            {
+                ruleName: "no-null/no-null",
```
JoshuaKGoldberg: This rule actually has a bit of a behavioral change from TSLint to ESLint. The ESLint rule doesn't handle `null` types:

```ts
let wat: string | null = null;
```

Could you please add a `notices` array explaining that?

da-snap:
I can't run the end-to-end tests. I have node 12.11.1 and tslint installed. It seems that the `testArgs.join(" ")` does not work for me.
When I try to log what is passed to exec I get something like this:
`node [...]/tslint-to-eslint-config/bin/tslint-to-eslint-config " - - c o n f i g "   "[...]/t s l i n t - t o - e s l i n t - c o n f i g / t e s t / t e s t s / s t a n d a l o n e   t s l i n t . j s o n / . e s l i n t r c . j s o n "   " - - t s l i n t "   " [...] / t s l i n t - t o - e s l i n t - c o n f i g / t e s t / t e s t s / s t a n d a l o n e   t s l i n t . j s o n / t s l i n t . j s o n "`

Do you have an Idea what could be the issue? Since it works in circleci...

JoshuaKGoldberg:
I'll take a look at the test shenanigans...

JoshuaKGoldberg:
Ignoring end-to-end test failures as they're removed in master. 🤷‍♂

PR #204, opened 2019-10-04
Add missing converter: file-name-casing

### LINKED ISSUE #161: Missing converter: file-name-casing
### 💥 Missing Converter

#### TSLint Rule Name

`file-name-casing`

da-snap:
Linked to issue: #161

## PR Checklist

-   [x] Addresses an existing issue: fixes #161

## Overview

Add converter for file-name-casing. Map casings to new rule in the unicorn plugin.

### DIFF
--- a/src/rules/converters/file-name-casing.ts
+++ b/src/rules/converters/file-name-casing.ts
@@ -1,6 +1,15 @@
 import { RuleConverter } from "../converter";
 
-const IGNORE_CASE_NOTICE = "ESLint (Unicorn plugin) does not support ignore as case";
+const IGNORE_CASE_NOTICE = "ESLint (Unicorn plugin) does not support the 'ignore' case.";
+const CASING_BY_FILETYPE_CHANGE =
+    "ESLint (Unicorn Plugin) does not support file name casing by file type, so all previously configured casings are now allowed.";
+const CASES_MAP: { [s: string]: string } = {
+    "camel-case": "camelCase",
+    "pascal-case": "pascalCase",
+    "kebab-case": "kebabCase",
+    "snake-case": "snakeCase",
+};
+
 export const convertFileNameCasing: RuleConverter = tslintRule => {
     return {
         rules: [
@@ -15,12 +24,7 @@ export const convertFileNameCasing: RuleConverter = tslintRule => {
 
 const collectArguments = (ruleArguments: any[]) => {
     const notices: string[] = [];
-    const casesMap = new Map();
-    casesMap.set("camel-case", "camelCase");
-    casesMap.set("pascal-case", "pascalCase");
-    casesMap.set("kebab-case", "kebabCase");
-    casesMap.set("snake-case", "snakeCase");
-    const foundCases: { [k: string]: any } = {};
+    const foundCases: { [k: string]: boolean } = {};
 
     if (ruleArguments.length === 0 || ruleArguments[0] === false || ruleArguments.length < 2) {
         return undefined;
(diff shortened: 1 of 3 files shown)


src/rules/converters/file-name-casing.ts
```diff
@@ -0,0 +1,56 @@
...
+    };
+};
+
+const collectArguments = (ruleArguments: any[]) => {
+    const notices: string[] = [];
+    const casesMap = new Map();
+    casesMap.set("camel-case", "camelCase");
+    casesMap.set("pascal-case", "pascalCase");
+    casesMap.set("kebab-case", "kebabCase");
+    casesMap.set("snake-case", "snakeCase");
```
JoshuaKGoldberg: You can make a map with initialized properties in the constructor:

```ts
new Map([
    ["camel-case", "camelCase"],
    // ...
]);
```

That being said, I don't think this really needs to be a `Map`. Only `.get` is used on it, and always for keys guaranteed to exist. Can you switch it to just a regular object, and initialize it outside of `collectArguments` (since it's only assigned to and never modified)?

src/rules/converters/file-name-casing.ts
```diff
@@ -0,0 +1,56 @@
+import { RuleConverter } from "../converter";
+
+const IGNORE_CASE_NOTICE = "ESLint (Unicorn plugin) does not support ignore as case";
```
JoshuaKGoldberg:
```suggestion
const IGNORE_CASE_NOTICE = "ESLint (Unicorn plugin) does not support the 'ignore' case.";
```

...to be a little more clear. Other `notices` end with periods too.

src/rules/converters/file-name-casing.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+const collectArguments = (ruleArguments: any[]) => {
+    const notices: string[] = [];
+    const casesMap = new Map();
+    casesMap.set("camel-case", "camelCase");
+    casesMap.set("pascal-case", "pascalCase");
+    casesMap.set("kebab-case", "kebabCase");
+    casesMap.set("snake-case", "snakeCase");
+    const foundCases: { [k: string]: any } = {};
```
JoshuaKGoldberg:
```suggestion
    const foundCases: { [k: string]: string } = {};
```

...right? Once `casesMap` is converted to a `{ [i: string]: string }`, this should work more easily in the typings.

src/rules/converters/tests/file-name-casing.test.ts
```diff
@@ -0,0 +1,106 @@
...
+                    ],
+                },
+            ],
+            plugins: ["unicorn"],
+        });
+    });
+
+    test("conversion with object for filetypes", () => {
+        const result = convertFileNameCasing({
+            ruleArguments: [true, { ".ts": "ignore", ".tsx": "pascal-case", ".js": "snake-case" }],
```
JoshuaKGoldberg: Aha, a behavior change! TSLint can specify file name casing per file but ESLint just lets all files have any case. There should be another `notices` member here for the difference.

da-snap:
Thank you for your feedback! I tried to fix all issues. Let me know if there is anything else.

src/rules/converters/tests/file-name-casing.test.ts
da-snap: I added a notice: `ESLint (Unicorn Plugin) does not support file name casing by file type, so all previously configured casings are now allowed.`

PR #222, opened 2019-10-06
Implement variable-name rule converter

### LINKED ISSUE #219: Missing converter: variable-name
### 💥 Missing Converter

-   `tslint-to-eslint-config` version: 0.2.7

#### TSLint Rule Name

`variable-name`

The issue here is that this package doesn't know how to convert the TSLint rule into the equivalent ESLint rule. The equivalent ESlint rule doesn't show on the [eslint-plugin Roadmap](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md).

KingDarBoja:
## PR Checklist

-   [x] Addresses an existing issue: fixes #219

## Overview

- Added a rule converter for `variable-name` in the TSLint ruleset.
- Added proper handling for the options that come with the rule itself.
- Added tests to make sure that the right comparative rules are being output.
- Added a couple of notices so that the user is informed of things not automatically converted.

This is my first PR so don't expect it to be 100% perfect so awaiting for any feedback from members/contributors.


src/rules/converters/variable-name.ts
```diff
@@ -0,0 +1,116 @@
...
+            );
+        }
+
+        if (
+            tslintRule.ruleArguments.length !== 0 &&
+            (tslintRule.ruleArguments.includes("allow-pascal-case") ||
+                tslintRule.ruleArguments.includes("allow-snake-case"))
+        ) {
+            camelCaseOptionNotice.push(
+                "This rule does not allow pascal neither snake case to variable names. Those are reserved for class names and static methods.",
```
JoshuaKGoldberg:
```suggestion
                "ESLint's camel-case rule does not allow pascal or snake case variable names. Those cases are reserved for class names and static methods.",
```

src/rules/converters/variable-name.ts
```diff
@@ -0,0 +1,116 @@
+import { RuleConverter } from "../converter";
+
+export const convertVariableName: RuleConverter = tslintRule => {
+    const getCamelCaseRuleOptions = () => {
+        const camelCaseOptionNotice: string[] = [];
+        camelCaseOptionNotice.push(
+            "By default, this rule looks for any underscores (_) located within the source code. It ignores leading and trailing underscores and only checks those in the middle of a variable name.",
```
JoshuaKGoldberg: Phrasing suggestion, since the notices are intended to inform the user of _differences_ between linter behaviors...

```suggestion
            "Leading and trailing underscores (_) in variable names will now be ignored.",
```

However, TSLint's `variable-name` rule has `allow-leading-underscore` and `allow-trailing-underscore` options, so there are a few cases for this notice:

* If both of those options are enabled, this notice shouldn't be added
* If one option is enabled, it should only say that type: e.g. _"Leading undescores in variable names ..."_
* If both options are enabled, this suggested message is :+1:

src/rules/converters/variable-name.ts
```diff
@@ -0,0 +1,116 @@
...
+            arguments: [],
+            notices: camelCaseOptionNotice,
+        };
+    };
+
+    const getUnderscoreDangleRuleOptions = () => {
+        const underscoreDangleOptionArguments: string[] = [];
+        const underscoreDangleOptionNotice: string[] = [];
+        underscoreDangleOptionNotice.push(
+            'By default, "no-underscore-dangle" will disallows dangling underscores in identifiers.',
```
JoshuaKGoldberg:
```suggestion
            'By default, "no-underscore-dangle" disallows dangling underscores in identifiers.',
```

src/rules/converters/variable-name.ts
```diff
@@ -0,0 +1,116 @@
...
+        return {
+            arguments: [],
+            notices: camelCaseOptionNotice,
+        };
+    };
+
+    const getUnderscoreDangleRuleOptions = () => {
+        const underscoreDangleOptionArguments: string[] = [];
+        const underscoreDangleOptionNotice: string[] = [];
+        underscoreDangleOptionNotice.push(
```
JoshuaKGoldberg: Instead of describing how the rule options have changed, you'll need to change this to describe how the end behavior has changed. See the other notice arrays for similar suggestions.

src/rules/converters/variable-name.ts
```diff
@@ -0,0 +1,116 @@
...
+    const camelCaseOptions = getCamelCaseRuleOptions();
+    const underscoreDangleOptions = getUnderscoreDangleRuleOptions();
+    const idblackListOptions = getBlackListRuleOptions();
+
+    return {
+        rules: [
+            {
+                ruleName: "camelcase",
+                ruleArguments: camelCaseOptions.arguments,
+                notices: camelCaseOptions.notices,
```
JoshuaKGoldberg: Here's a fun way to save a line and IMO make the code more clear:

```suggestion
               ...camelCaseOptions,
```

...and since the variable is only used once, you can get this a bit more terse with:

```ts
{
   ruleName: "camelcase",
   ...getCamelCaseRuleOptions(),
},

...if you change the objects in the `get*Options()` methods to return `ruleArguments` instead of `arguments`.
```

src/rules/converters/variable-name.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export const convertVariableName: RuleConverter = tslintRule => {
+    const getCamelCaseRuleOptions = () => {
+        const camelCaseOptionNotice: string[] = [];
+        camelCaseOptionNotice.push(
+            "By default, this rule looks for any underscores (_) located within the source code. It ignores leading and trailing underscores and only checks those in the middle of a variable name.",
+        );
+
+        if (
+            tslintRule.ruleArguments.length !== 0 &&
```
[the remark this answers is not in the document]
KingDarBoja: You are right, however I tend to check first the array length to avoid any extra lookup on the array itself. I don't know where I got that habit but nice to hear that suggestion :)

src/rules/converters/variable-name.ts
```diff
@@ -0,0 +1,116 @@
...
+                "number",
+                "String",
+                "string",
+                "Boolean",
+                "boolean",
+                "Undefined",
+                "undefined",
+            );
+            blackListOptionNotice.push(
+                'If "ban-keywords" was provided, ESLint has to disallows the use of certain TypeScript keywords by using "id-blacklist" rule.',
```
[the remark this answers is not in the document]
KingDarBoja: Ah! I wasn't sure about how notices works on rule conversion, my first thought was describing rule details of why those options were added from TSLint into ESLint.

src/rules/converters/variable-name.ts
KingDarBoja: Oh nice, the object destructuring behaviour, not used to it but glad to hear the suggestion.

src/rules/converters/variable-name.ts
KingDarBoja: The notice should be displayed if **both** options are enabled or not?

JoshuaKGoldberg: If both options are enabled the notice should not be displayed. ESLint's rule behaves the same as the TSLint rule with both options enabled (right?).

src/rules/converters/variable-name.ts
```diff
@@ -0,0 +1,121 @@
...
+            ) {
+                camelCaseOptionNotice.push(
+                    "Trailing undescores in variable names will now be ignored.",
+                );
+            }
+        } else {
+            camelCaseOptionNotice.push(
+                "Leading and trailing underscores (_) in variable names will now be ignored.",
+            );
+        }
```
JoshuaKGoldberg: I'm finding this section a little difficult to read... Would you mind changing it to something like:

```ts
const allowedLeadingUnderscore = tslintRule.ruleArguments.includes("allow-leading-underscore");
const allowedTrailingUnderscore = tslintRule.tslintRuleArguments.includes("allow-trailing-underscore");

if (allowedLeadingUnderscore) {
    if (!allowedTrailingUnderscore) {
        camelCaseOptionNotice.push("...");
    }
} else if (allowedTrailingUnderscore) {
     // ...
} else {
     // ...
}
```

_(and similar changes for later functions, if they end up being smaller this way?)_

src/rules/converters/variable-name.ts
```diff
@@ -0,0 +1,121 @@
...
+                ...getUnderscoreDangleRuleOptions(),
+            },
+            {
+                ruleName: "id-blacklist",
+                ...getBlackListRuleOptions(),
+            },
+            {
+                ruleName: "id-match",
+            },
+        ],
```
JoshuaKGoldberg: Sorry, _one_ more LoC optimization here 😅: if you move `ruleName`s to the `get*RuleOptions` methods, this whole array becomes:

```ts
rules: [
    getCamelCaseRuleOptions(),
    getUnderscoreDangleRuleOptions(),
    getBlackListRuleOptions(),
    {
        ruleName: "id-match",
    },
],
```

src/rules/converters/variable-name.ts
KingDarBoja: YES! Can do it :) I do get confused also with those nested if-else statements, was thinking on using switch case instead.

PR #223, opened 2019-10-06
Implement no-submodule-imports converter

### LINKED ISSUE #175: Missing converter: no-submodule-imports
### 💥 Missing Converter

#### TSLint Rule Name

`no-submodule-imports`

KingDarBoja:
## PR Checklist

-   [x] Addresses an existing issue: fixes #175

## Overview

Add converter based on the rule definition at TSLint, also it requires a plugin to be installed.


src/rules/converters/no-submodule-imports.ts
```diff
@@ -0,0 +1,28 @@
...
+    const ruleArguments: string[] = [];
+
+    if (
+        !(
+            tslintRule.ruleArguments.length === 0 ||
+            tslintRule.ruleArguments[0] === false ||
+            tslintRule.ruleArguments.length < 2
+        )
+    ) {
+        ruleArguments.push(...tslintRule.ruleArguments);
```
JoshuaKGoldberg: The [TSLint `no-submodule-imports` docs](https://palantir.github.io/tslint/rules/no-submodule-imports/) state that you can provide a package name like `rxjs` for the whole package to be allowed. The [ESLint `no-internal-modules` docs](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-internal-modules.md) don't have an equivalent. Can you confirm that ESLint will allow all `rxjs` modules in this case? If not, I think you'll need to convert it to `rxjs/*`, right?

KingDarBoja: Yeah, the `ESLint` docs states that it should be a minimatch/glob patterns instead of just the module name.

I will check out if it works like that instead of using the glob pattern.

KingDarBoja: Argh, you are right, we have to provide those keys as glob patterns, so instead of only `rxjs`, we should do `rxjs/*`.

I will submit the changes as soon as possible.

src/rules/converters/tests/no-submodule-imports.test.ts
```diff
@@ -0,0 +1,34 @@
+import { convertNoSubmoduleImports } from "../no-submodule-imports";
+
+describe(convertNoSubmoduleImports, () => {
```
JoshuaKGoldberg: Going off the TSLint rule docs, there should be a test case for a rule arguments list like `["@angular/platform-browser"]`.

KingDarBoja:
> 👌

That's awesome! Also can we remove the `waiting for author` label? 🍰

PR #230, opened 2019-10-07
Added missing no-implicit-dependencies converter

### LINKED ISSUE #171: Missing converter: no-implicit-dependencies
### 💥 Missing Converter

#### TSLint Rule Name

`no-implicit-dependencies`

melbarch:
## PR Checklist

-   [x] Addresses an existing issue: fixes #171 

## Overview
Adds a missing rule converter for the no-implicit-dependencies rule.

Still not sure how to handle correctly the case when whitelist modules are given :
I saw [here](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-extraneous-dependencies.js) that all (dev dep, peer dep, opt dep ...) accepts the whitelist array, but I added them only on devDep for the moment, as they are not mapped to specific category of deps on Tslint side.

Any thoughts on this @JoshuaKGoldberg 👋

### DIFF
--- a/src/rules/converters/no-implicit-dependencies.ts
+++ b/src/rules/converters/no-implicit-dependencies.ts
@@ -8,6 +8,8 @@ export const convertNoImplicitDependencies: RuleConverter = tslintRule => {
             ruleArgs.push({ devDependencies: false });
         } else if (element === "optional") {
             ruleArgs.push({ optionalDependencies: false });
+        } else {
+            ruleArgs.push({ devDependencies: element });
         }
     });
     return {
--- a/src/rules/converters/no-implicit-dependencies.ts
+++ b/src/rules/converters/no-implicit-dependencies.ts
@@ -1,7 +1,10 @@
 import { RuleConverter } from "../converter";
 
 export const convertNoImplicitDependencies: RuleConverter = tslintRule => {
-    const ruleArgs = [];
+    const ruleArgs: {
+        [eslintOption in "devDependencies" | "optionalDependencies"]?: boolean
+    }[] = [];
+    const notices: string[] = [];
 
     tslintRule.ruleArguments.forEach(element => {
         if (element === "dev") {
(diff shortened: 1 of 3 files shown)


src/rules/converters/no-implicit-dependencies.ts
```diff
@@ -0,0 +1,23 @@
+import { RuleConverter } from "../converter";
+
+export const convertNoImplicitDependencies: RuleConverter = tslintRule => {
+    const ruleArgs = [];
```
KingDarBoja: This is causing a error on the CI test [as seen here](https://circleci.com/gh/typescript-eslint/tslint-to-eslint-config/2105?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link).

> _"TS7034: Variable 'ruleArgs' implicitly has type 'any[]' in some locations where its type cannot be determined."_

src/rules/converters/no-implicit-dependencies.ts
```diff
@@ -0,0 +1,23 @@
...
+            ruleArgs.push({ devDependencies: element });
+        }
+    });
+    return {
+        rules: [
+            {
+                ruleName: "import/no-extraneous-dependencies",
+                ...(ruleArgs.length > 0 && { ruleArguments: ruleArgs }),
+            },
+        ],
```
KingDarBoja: I am pretty sure you need to provide the plugin after rules array as it is being handled by a plugin. Check [eslint plugin Roadmap](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md) for more info.

src/rules/converters/no-implicit-dependencies.ts
melbarch: Thanks, i'm looking into it

JoshuaKGoldberg:
Yeah @melbarch I think the configurations are inherently incompatible. TSLint allows you to filter out _module names_, while ESLint allows you to filter out _files that are importing them_.

The resolution for tslint-to-eslint-config will be to give up including the TSLint whitelist option and add a 'notice' complaining about the difference. 🤷‍♂

PR #237, opened 2019-10-12
Add strict-boolean-expression converter

### LINKED ISSUE #185: Missing converter: strict-boolean-expressions
### 💥 Missing Converter

#### TSLint Rule Name

`strict-boolean-expressions`

KingDarBoja:
## PR Checklist

-   [x] Addresses an existing issue: fixes #185

## Overview

- Added a rule converter for `stric-boolean-expressions` in the TSLint ruleset.
- Added proper handling for the options that come with the rule itself.
- Added tests to make sure that the right comparative rules are being output.
- Added a notice so that the user is informed of things not automatically converted or covered.
- Remove comments at `converters` file of already solved converters.

### DIFF
--- a/src/rules/converters/strict-boolean-expressions.ts
+++ b/src/rules/converters/strict-boolean-expressions.ts
@@ -1,0 +1,37 @@
+import { RuleConverter } from "../converter";
+
+export const ForbiddenOtherNonBooleanTypes =
+    "String, number, enum and mixed union types are forbidden.";
+
+export const convertStrictBooleanExpressions: RuleConverter = tslintRule => {
+    const notices: string[] = [];
+    const ruleArguments: { [key: string]: boolean } = {};
+
+    if (tslintRule.ruleArguments.length >= 2 || tslintRule.ruleArguments[0] === true) {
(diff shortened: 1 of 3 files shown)


src/rules/converters/strict-boolean-expressions.ts
```diff
@@ -0,0 +1,37 @@
+import { RuleConverter } from "../converter";
+
+export const ForbiddenOtherNonBooleanTypes =
+    "String, number, enum and mixed union types are forbidden.";
```
JoshuaKGoldberg: It would be a little confusing to get this general error if you didn't already have some of those options enabled. It should really be specific to the input rule arguments.

For example, if you allowed number, enum, and mixed union types, you should probably get:

```
String type comparisons are now forbidden.
```

KingDarBoja: The problem is that neither of those types are allowed, no matter if you set them on the tslint rule. The eslint doesn't allow it, only the `null` and `undefined` types are.

JoshuaKGoldberg: Ok dokes, you've convinced me. 😄

PR #239, opened 2019-10-14
add no-import-side-effect converter

### LINKED ISSUE #172: Missing converter: no-import-side-effect
### 💥 Missing Converter

#### TSLint Rule Name

`no-import-side-effect
`

bmeverett:
## PR Checklist

-   [x] Addresses an existing issue: fixes #172

## Overview

I added the converter to convert the rule to ESLint. I didn’t try and convert the reflex to glob syntax. I can explore that if you feel it would be valuable.

### DIFF
--- a/src/rules/converters/no-import-side-effect.ts
+++ b/src/rules/converters/no-import-side-effect.ts
@@ -1,11 +1,9 @@
 import { RuleConverter } from "../converter";
 
 export const convertNoImportSideEffect: RuleConverter = tsLintRule => {
-    const rules = [];
     const notices = [];
 
     if (tsLintRule.ruleArguments.length > 0) {
-        rules.push({ allow: tsLintRule.ruleArguments[1]["ignore-module"] });
         notices.push(
             "ESLint's no-import-side-effect now accepts a glob pattern for ignores; you'll need to manually convert your ignore-module settings.",
         );
@@ -14,7 +12,7 @@ export const convertNoImportSideEffect: RuleConverter = tsLintRule => {
     return {
         rules: [
             {
-                ruleArguments: rules,
+                ruleArguments: [],
                 ruleName: "no-import-side-effect",
                 notices: notices,
             },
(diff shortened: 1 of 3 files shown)


src/rules/converters/no-import-side-effect.ts
```diff
@@ -0,0 +1,21 @@
+import { RuleConverter } from "../converter";
+
+export const convertNoImportSideEffect: RuleConverter = tsLintRule => {
+    const rules = [];
+    const notices = [];
+
+    if (tsLintRule.ruleArguments.length > 0) {
+        rules.push({ allow: tsLintRule.ruleArguments[1]["ignore-module"] });
+        notices.push("This now accepts a glob pattern, so your regex may need converted");
```
JoshuaKGoldberg: What do you think of slight phrasing change?

```suggestion
        notices.push("ESLint's no-import-side-effect now accepts a glob pattern for ignores; you'll need to manually convert your ignore-module settings.");
```

src/rules/converters/no-import-side-effect.ts
```diff
@@ -0,0 +1,23 @@
+import { RuleConverter } from "../converter";
+
+export const convertNoImportSideEffect: RuleConverter = tsLintRule => {
+    const rules = [];
+    const notices = [];
+
+    if (tsLintRule.ruleArguments.length > 0) {
+        rules.push({ allow: tsLintRule.ruleArguments[1]["ignore-module"] });
```
JoshuaKGoldberg: Oh, I just realized this is directly copying the original `ignore-module` setting over... but, as the notice mentions, they're different formats. It could be confusing to see the original regex placed in the ESLint rule options, and potentially a cause of unexpected behavior differences. IMO it'd be better to just not add an `allow` argument at all.

Sorry for not commenting on this sooner!

bmeverett: No worries, I'll take a look

buu700:
@bmeverett it doesn't look like eslint has a `no-import-side-effect` rule. Was this a mistake?

JoshuaKGoldberg:
🤦‍♂ oops, I should have checked that. Thank you @buu700! Fixed in #271.

PR #240, opened 2019-10-16
Add comment-format converter

KingDarBoja:
## PR Checklist

-   [ ] Addresses an existing issue

## Overview

Add missing `comment-format` converter as already being listed at [ROADMAP](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md)

### DIFF
--- a/src/rules/converters/comment-format.ts
+++ b/src/rules/converters/comment-format.ts
@@ -5,7 +5,6 @@ type CommentFormatOptions = {
     "ignore-pattern": string;
 };
 
-export const CheckTrailingLowercaseMessage = "Only first trailing comment can be configured.";
 export const CapitalizedIgnoreMessage = "Only accepts a single string pattern to be ignored.";
 
 export const convertCommentFormat: RuleConverter = tslintRule => {
@@ -16,7 +15,6 @@ export const convertCommentFormat: RuleConverter = tslintRule => {
     const hasCheckSpace = tslintRule.ruleArguments.includes("check-space");
     const hasCheckLowercase = tslintRule.ruleArguments.includes("check-lowercase");
     const hasCheckUppercase = tslintRule.ruleArguments.includes("check-uppercase");
-    const hasCheckTrailingLowercase = tslintRule.ruleArguments.includes("allow-trailing-lowercase");
 
     if (!hasCheckSpace) {
         spaceCommentRuleArguments.push("never");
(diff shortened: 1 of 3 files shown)


src/rules/converters/comment-format.ts
```diff
@@ -0,0 +1,56 @@
...
+        capitalizedRuleArguments.push("always");
+    } else if (hasCheckLowercase) {
+        capitalizedRuleArguments.push("never");
+    }
+
+    if (hasCheckTrailingLowercase) {
+        capitalizedNotices.push(CheckTrailingLowercaseMessage);
+    }
+
+    if (typeof tslintRule.ruleArguments[tslintRule.ruleArguments.length - 1] == "object") {
```
JoshuaKGoldberg: Huh, I'm surprised this isn't enabled in the lint ruleset already...

```suggestion
    if (typeof tslintRule.ruleArguments[tslintRule.ruleArguments.length - 1] === "object") {
```

src/rules/converters/comment-format.ts
```diff
@@ -0,0 +1,56 @@
...
+        const objectArgument: Object =
```
JoshuaKGoldberg: Could you make a type for this? `Object` is a pretty vague type.

src/rules/converters/comment-format.ts
```diff
@@ -0,0 +1,56 @@
...
+        spaceCommentRuleArguments.push("never");
+    }
+
+    if (hasCheckUppercase) {
+        capitalizedRuleArguments.push("always");
+    } else if (hasCheckLowercase) {
+        capitalizedRuleArguments.push("never");
+    }
+
+    if (hasCheckTrailingLowercase) {
```
JoshuaKGoldberg: Doesn't ESLint's `ignoreConsecutiveComments` satisfy this use case?

src/rules/converters/comment-format.ts
KingDarBoja: The triple equal lint? Yeah, I didn't saw any warning/error message after doing the type check.

src/rules/converters/comment-format.ts
KingDarBoja: Of course! Just wait a minute 👍

src/rules/converters/comment-format.ts
KingDarBoja: You are right! Didn't noticed about it earlier.

JoshuaKGoldberg: I'm interpreting them to have the same intent: that only the first in a series of comments will be linted. ESLint's wording is a little confusing here.

KingDarBoja: @JoshuaKGoldberg it is done, I removed it.

JoshuaKGoldberg:
Just waiting on the `test` task to complete - missing coverage on line 32 of `comment-format.ts`. :+1:

KingDarBoja:
> Just waiting on the `test` task to be fixed - missing coverage on line 32 of `comment-format.ts`. 👍

I already covered that line but not sure why the test coverage is saying it doesn't 😢

JoshuaKGoldberg:
Ahh, it's _partially_ covered. Branch coverage is <100% so although the line is hit, not every case in it is. Expanding in `file:///C:/Code/tslint-to-eslint-config/coverage/lcov-report/src/rules/converters/comment-format.ts.html` _(or whatever your file path equivalent would be)_:

image

I think the missing case is `objectArgument["ignore-words"].length` being `0`?

KingDarBoja:
@JoshuaKGoldberg Good catch! Now it is ready 🍰

PR #245, opened 2019-10-20
Factor in 'raw' configurations on top of 'reported' ones

### LINKED ISSUE #232: Missing extends in generated config
### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 0.2.7
-   ESLint version: 6.5.1
-   Node version: 12.10.0

#### Actual Behavior

`extends` is an empty array in the output even if there were `extends` in the original `.eslintrc`

#### Expected Behavior

The `extends` in the original `.eslintrc` should be retained

#### Reproduction

Original `.eslintrc.js`:

```js
module.exports = {
  extends: ["airbnb"]
}
```

Output using ESLint 5:
```js
module.exports = {
    "env": {
        "browser": true,
        "es6": true
    },
    "extends": [
        ".../node_modules/eslint-config-airbnb-base/rules/best-practices.js",
        ".../node_modules/eslint-config-airbnb-base/rules/errors.js",
        ".../node_modules/eslint-config-airbnb-base/rules/node.js",
        ".../node_modules/eslint-config-airbnb-base/rules/style.js",
        ".../node_modules/eslint-config-airbnb-base/rules/variables.js",
        ".../node_modules/eslint-config-airbnb-base/rules/es6.js",
        ".../node_modules/eslint-config-airbnb-base/rules/imports.js",
        ".../node_modules/eslint-config-airbnb-base/rules/strict.js",
        ".../node_modules/eslint-config-airbnb-base/index.js",
        ".../node_modules/eslint-config-airbnb/rules/react.js",
        ".../node_modules/eslint-config-airbnb/rules/react-a11y.js",
        "airbnb"
    ],
    ...
};
```
Output using ESLint 6:
```js
module.exports = {
    "env": {
        "browser": true,
        "es6": true
    },
    "extends": [],
    ...
};
```

It seems that the behaviour of the `--print-config` flag changed in ESLint 6 and it no longer prints `extends`

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #232

## Overview

Before this change, `find*Configuration` functions for linters only returned the 'reported' config, meaning the CLI output of `--print-config` or similar. This change makes them return two things:

* `full`: that reported configuration, considered to be the complete list of all original settings
* `raw`: the naive results of `import`ing the configuration file

The `raw` config is used to override some settings in the printed output, namely `globals` and `extends`, because ESLint prints _way_ too many globals but none of the extends (#232).

A few tangential changes:

* A new `importer` was necessary for importing `.json` files with comments.
* Added root object key sorting to the output formatters, since they were getting a little oddly sorted from the raw overrides
* Enabled `esModuleInterop` in `tsconfig.json` because `strip-json-comments`' types don't have a default export

Still to do:

- [x] Flesh out test coverage, especially for the new `importer`
- [x] Add some JSDoc comments around the more confusing areas... perhaps a `.md` addition?
- [x] Double-check this works locally _(initial runs worked well!)_

### DIFF
--- a/src/input/findTSLintConfiguration.ts
+++ b/src/input/findTSLintConfiguration.ts
@@ -1,44 +1,67 @@
+import { findRawConfiguration } from "./findRawConfiguration";
+import { findReportedConfiguration } from "./findReportedConfiguration";
 import { Exec } from "../adapters/exec";
-import { findConfiguration } from "./findConfiguration";
+import { OriginalConfigurations } from "./findOriginalConfigurations";
+import { SansDependencies } from "../binding";
+import { importer } from "./importer";
 
 export type TSLintConfiguration = {
+    extends?: string[];
     rulesDirectory: string[];
     rules: TSLintConfigurationRules;
 };
 
-export type TSLintConfigurationRules = {
-    [i: string]: any;
-};
+export type TSLintConfigurationRules = Record<string, any>;
 
 const defaultTSLintConfiguration = {
+    extends: [],
     rulesDirectory: [],
     rules: {},
 };
 
 export type FindTSLintConfigurationDependencies = {
     exec: Exec;
+    importer: SansDependencies<typeof importer>;
 };
 
 export const findTSLintConfiguration = async (
     dependencies: FindTSLintConfigurationDependencies,
     config: string | undefined,
-): Promise<TSLintConfiguration | Error> => {
-    const rawConfiguration = await findConfiguration<TSLintConfiguration>(
-        dependencies.exec,
-        "tslint --print-config",
-        config || "./tslint.json",
-    );
+): Promise<OriginalConfigurations<TSLintConfiguration> | Error> => {
+    const filePath = config || "./tslint.json";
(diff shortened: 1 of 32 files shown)


src/input/findTSLintConfiguration.ts
```diff
@@ -1,44 +1,67 @@
...
+    if (rawConfiguration instanceof Error) {
         return rawConfiguration;
     }
 
     return {
-        ...defaultTSLintConfiguration,
-        ...rawConfiguration,
+        full: {
+            ...defaultTSLintConfiguration,
+            ...rawConfiguration,
```
jeremyyap: Hi @JoshuaKGoldberg I noticed TSLint rule arguments are not being picked up in the latest release and I think the issue might be here. Should this be `reportedConfiguration` instead?

JoshuaKGoldberg: Perhaps! @jeremyyap would you mind filing an issue! I can take a look later today.

PR #250, opened 2019-10-21
Feature/convert editor settings 133

### LINKED ISSUE #133: Automatic vscode settings migration
> why you no .vscode settings? 😒

### 🚀 Feature Request

It would be awesome if this tool could migrate .vscode settings from tslint to eslint. For example:

Change this:

```js
"editor.codeActionsOnSave": {
    "source.fixAll.tslint": true
}
```

To:

```js
"eslint.autoFixOnSave": true
```

Vscode settings can reside in either `.vscode/settings.json` or a `*.vscode-workspace` file.

#### Existing Behavior

No vscode support, bro. 🥺

#### Change Proposal

Probably this should be under a compiler flag.

Also, thanks for this awesome tool 👏, I will check it out later when I have some time.

MrCube42:
## PR Checklist

-   [x] Addresses an existing issue: fixes #133

## Overview

- Added CLI option `--editor` to specify an editor config path, and have it default to `.vscode/settings.json`
- Renamed the method passed to `runCli.ts` into `convertConfigs` (plural)
- Change `runCli` to do several conversions after each other
- Added functionality to read VSCode **editor configuration**
- Added functionality to convert VSCode **editor settings** from this configuration describe in issue #133 
```
"editor.codeActionsOnSave": {
    "source.fixAll.tslint": true
}
```
To:
```
"eslint.autoFixOnSave": true
```


src/conversion/convertEditorConfig.ts
```diff
@@ -0,0 +1,79 @@
...
+        ...originalEditorConfiguration,
+        "editor.codeActionsOnSave": codeActionsOnSaveWithoutReplacedProperties,
+        "eslint.autoFixOnSave": originalSourceFixAllTsLint,
+    };
+
+    const output = newEditorConfiguration;
+    const outputPath = settings.editor ? settings.editor : DEFAULT_VSCODE_SETTINGS_PATH;
+
+    try {
+        await dependencies.fileSystem.writeFile(outputPath, formatOutput(outputPath, output));
```
JoshuaKGoldberg: This `writeFile` line won't throw, so you could do this instead of the `try`/`catch`:

```suggestion
    const fileWriteError = await dependencies.fileSystem.writeFile(outputPath, formatOutput(outputPath, output));
    if (fileWriteError !== undefined) {
        // ...
```

src/conversion/convertEditorConfig.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    // 4. Rewrite editor configuration with replacements.
+    const newEditorConfiguration = {
+        ...originalEditorConfiguration,
+        "editor.codeActionsOnSave": codeActionsOnSaveWithoutReplacedProperties,
+        "eslint.autoFixOnSave": originalSourceFixAllTsLint,
+    };
+
+    const output = newEditorConfiguration;
+    const outputPath = settings.editor ? settings.editor : DEFAULT_VSCODE_SETTINGS_PATH;
```
JoshuaKGoldberg: Mmhh, yeah, this looks like it should be moved to a different function that's passed as a part of `dependencies` to here. It'd be hard to test as-is here (reasonably so - this function is pretty big). I think that's what you meant by the comments?

src/cli/runCli.ts
```diff
@@ -36,15 +42,13 @@ export const runCli = async (
...
 
-    try {
-        result = await dependencies.convertConfig(parsedArgv);
-    } catch (error) {
-        result = {
-            errors: [error as Error],
-            status: ResultStatus.Failed,
-        };
+    for (const convertConfig of dependencies.convertConfigs) {
+        result = await tryConvertConfig(convertConfig, parsedArgv);
```
JoshuaKGoldberg: Reassigning `result` means if the first one fails you'll end up not reporting that error. Unfortunate. How about pushing to `result.errors` if need be, and having `status` be whether there are multiple errors?

src/conversion/convertEditorConfig.ts
devinmotion: Absolutely true. Changed it.

src/cli/runCli.ts
devinmotion: Reporting the errors to console and immediately return if failed, otherwise succeed.

src/conversion/convertEditorConfig.ts
devinmotion: Will work on this. Should work the same as convertRules in convertConfig.

devinmotion: @JoshuaKGoldberg I splitted readingConfig, conversion (with extensible settings convertions), and writing of the results like it is done for the rules. I don't report any conversions yet. Tests have so be written. Could you please have a look at my new approach?

JoshuaKGoldberg: Flagging this to look at it soon!

JoshuaKGoldberg: Sorry for the delay @devinmotion! i wanted to give this a good reading-through.

Yes, this looks like a great approach. I think it'll work quite well, and it looks really polished already. Awesome! 🚀 

A few notes:

* `reportSettingConversionResults` seems like a subset of `reportConversionResults`. I wonder if you could have a single function that receives a standardized input object and logs the output in a similar way between them? Seems like you'd remove a good portion of copypasta from that particular new file.
* Similarly, `convertConfig` and `convertEditorConfig` are _really_ similar... I wonder if the common parts from both should be moved to some common `runFileConversion` helper? 🤷‍♂ maybe better to file a followup ticket to investigate; that feels like a kind of long-winded architectural debate.
* Reading through the branch, the difference between general linter config names (e.g. `convertConfig`) and their settings file equivalents (e.g. `convertSettingsConfig`) was a little hard to keep track of. That seems like something good for a followup issue+PR as well to clarify though, to keep the PR from getting too massive...

devinmotion: No problem. I really appreciate your extensive feedback. It is really helpful for me. 🙏 

As soon as I will find some time, I'll do the following to get the PR ready:
- Refactor `reportSettingConversionResults` and `reportConversionResults` to keep it more DRY
- Try to refactor `convertConfig` and `convertEditorConfig` to be more DRY. If this looks too costly, I'll rather create a followup issue
- Create a followup issue to improve on the `convertConfig` and `convertSettingsConfig` similarities and differences
- Write tests for new features

devinmotion: Refactor reportSettingConversionResults and reportConversionResults to keep it more DRY ✔️ including tests

devinmotion: Try to refactor convertConfig and convertEditorConfig to be more DRY. If this looks too costly, I'll rather create a followup issue ℹ️

Tried to do some refactorings but could not get that far because it seems to be complex without further changes of how rules and editor settings are converted.
Therefore I created a follow-up issue: https://github.com/typescript-eslint/tslint-to-eslint-config/issues/274

devinmotion: Create a followup issue to improve on the convertConfig and convertSettingsConfig similarities and differences ✔️ 
Issue created, needs clarification first: https://github.com/typescript-eslint/tslint-to-eslint-config/issues/275

devinmotion: 🚧 Working on the last missing tests:
- `convertSettings.ts`
- `writeEditorConfigConversionResults.ts`
- `convertEditorConfig.ts`

devinmotion:
@JoshuaKGoldberg Finally I managed to write the missing unit tests. As commented before, I created 2 follow up issues (#274, #275) to refactor and clean up some similar structures of rules and settings conversion. #275 should also address the general naming scheme of *editor setting configurations* in source code.

I am really sorry that this PR escalated much 😳 I was surprised myself that this "small" feature involved so many changes; quite a though one for "good first issue" 😉

If you find the time, just let me know if you want me to do something more in this PR.

Last but not least thank you for your really good feedback and your time to review my work. 🙏

KingDarBoja:
@devinmotion Wow man, you really nailed it! 🚀

JoshuaKGoldberg:
😮 indeed, not a 'good' first issue - removed the label from the issue! 🥇 to you for sticking through this PR! I'm reviewing now.

src/reporting/reportOutputs.ts
```diff
@@ -0,0 +1,74 @@
+import chalk from "chalk";
+import { EOL } from "os";
+
+import { Logger } from "../adapters/logger";
+import { ErrorSummary } from "../errors/errorSummary";
+
+export const logSuccessfulConversions = <K, V>(
```
JoshuaKGoldberg: It's not clear what generics in this file such as `K`_(ey)_ and `V`_(alue)_ stand for without reading through the functions. Could you switch to more informative names for them?

src/cli/main.ts
```diff
@@ -27,22 +32,28 @@ import { findPackagesConfiguration } from "../input/findPackagesConfiguration";
...
-import {
-    reportConversionResults,
-    ReportConversionResultsDependencies,
-} from "../reporting/reportConversionResults";
-import { converters } from "../rules/converters";
+import { reportSettingConversionResults } from "../reporting/reportSettingConversionResults";
+import { reportConversionResults } from "../reporting/reportConversionResults";
+import { ReportConversionResultsDependencies } from "../reporting/dependencies";
+import { converters as rulesConverters } from "../rules/converters";
+import { converters as settingsConverters } from "../settings/converters";
```
JoshuaKGoldberg: I'm finding it hard to keep track of multiple `converters` in dependencies. What do you think of renaming them to always be referred to as `rulesConverters` and `settingsConverters`?

src/settings/convertSettings.ts
```diff
@@ -0,0 +1,48 @@
...
+    const failed: ConversionError[] = [];
+    const missing: Pick<EditorSetting, "settingName">[] = [];
+
+    for (const [settingName, value] of Object.entries(rawEditorSettings)) {
+        const editorSetting = { settingName, value };
+        const conversion = convertSetting(editorSetting, dependencies.converters);
+
+        if (conversion === undefined) {
+            const { settingName } = editorSetting;
+            missing.push({ settingName });
```
JoshuaKGoldberg: Wait, I don't think we want to be logging all these ones as `missing`. A log file from TSLint's `.vscode/settings.json` gives these settings complaints:

```
editor.insertSpaces does not yet have an ESLint equivalent.
editor.tabSize does not yet have an ESLint equivalent.
editor.wordWrap does not yet have an ESLint equivalent.
editor.wordWrapColumn does not yet have an ESLint equivalent.
editor.wrappingIndent does not yet have an ESLint equivalent.
files.eol does not yet have an ESLint equivalent.
search.exclude does not yet have an ESLint equivalent.
typescript.tsdk does not yet have an ESLint equivalent.
```

It's really only the editor settings in https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-tslint-plugin#configuration that need to be mentioned, right?

src/settings/convertSettings.ts
```diff
@@ -0,0 +1,48 @@
...
+            continue;
+        }
+
+        if (conversion instanceof ConversionError) {
+            failed.push(conversion);
+            continue;
+        }
+
+        for (const changes of conversion.settings) {
+            converted.set(settingName, { ...changes });
```
JoshuaKGoldberg: When I try out this config locally:

```json
{
    "editor.codeActionsOnSave": {
        "source.fixAll.tslint": true
    }
}
```

...the two output rule settings clobber each other under the same key:

```json
{
    "editor.codeActionsOnSave": true
}
```

Did you mean this?:

```suggestion
            converted.set(changes.settingName, { ...changes });
```

src/input/findEditorConfiguration.ts
```diff
@@ -0,0 +1,35 @@
...
+const readConfiguration = async <Configuration>(
+    fileSystem: Pick<FileSystem, "readFile">,
+    path: string,
+): Promise<DeepPartial<Configuration> | Error> => {
+    try {
+        const fileContents = await fileSystem.readFile(path);
+        if (fileContents instanceof Error) {
+            throw new Error(fileContents.message);
+        }
+        return JSON.parse(fileContents) as DeepPartial<Configuration>;
```
JoshuaKGoldberg: Ah, but many editor settings include comments -- even including [TSLint's](https://github.com/palantir/tslint/blob/master/.vscode/settings.json). This PR crashes as is for technically invalid JSON such as `//` comments. 

`findRawConfiguration` gets around it by using `import` (via `importer`). That should work for _importing_ the contents... but then you'd get the issue of comments being removed. That seems "fine" as a followup issue; adding (or finding) some kind of fancy diff-based output (or using TypeScript's?!) seems... hard.

src/settings/convertSetting.ts
```diff
@@ -0,0 +1,19 @@
+import { ConversionError } from "../errors/conversionError";
+import { EditorSettingConverter } from "./converter";
+import { EditorSetting } from "./types";
+
+export const convertSetting = (
```
JoshuaKGoldberg: Yeah, coming back to this, I think this should really be called `convertEditorSetting`. It's a bit confusing at first read!

src/reporting/reportOutputs.ts
MrCube42: Replace with actual types `<string, EditorSetting | ESLintRuleOptions>`.

src/cli/main.ts
MrCube42: Renamed files and exported converters, so that we can import and use them directly.

src/settings/convertSettings.ts
MrCube42: That's true. Previously tried to convert the entire **editor configuration** (`.vscode/settings.json`). Changed it to handle only the **editor settings** from the **editor configuration** (those starting with `editor.`).

This change also indicates the naming as mentioned below: The **editor configuration** is the entire file like `.vscode/settings` while the **editor setting** are only those keys in the config that start with `editor.`.

src/settings/convertSettings.ts
MrCube42: Indeed! Thanks for the catch. Unfortunately I wrote an erroneous test that tested exactly the wrong case 🤦‍♂

src/input/findEditorConfiguration.ts
MrCube42: Changed it to use `findRawConfiguration`. Works well as described and is less custom code. Thanks.

src/settings/convertSetting.ts
MrCube42: Refactored and specified it as follows:
- **editorConfiguration** is the entire configuration file, e.g. `.vscode/settings.json`. Therefore `findEditorConfiguration` and `writeEditorConfigConversionResults` handle the entire config
- **editorSettings** are entries of the editor configuration prefixed with `editor.`, thus `convertEditorSettings` and `reportEditorSettingConversionResults` only handle this subset

MrCube42: Also renamed the property `settingName` in `EditorSetting` to `editorSettingName`. However, I am a bit unsure whether `settingName` would suffice.
@JoshuaKGoldberg What do you think?

JoshuaKGoldberg: Heh, and we could probably switch those names and still see it as all making sense... For now, I think it looks nice as is - erring on the side of caution given how many areas have confusingly similar names. Unless you start to feel strongly about going to `settingName`, let's leave it as `editorSettingName`? 🤷‍♂

PR #273, opened 2019-11-26
Generating `space-before-function-paren` (#147)

### LINKED ISSUE #147: Incorrectly generated space-before-function-paren
### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 0.2.5
-   ESLint version: 6.2.2
-   Node version: 12.6

#### Actual Behavior

The tool generates a

 ```json
        "space-before-function-paren": [
            "error",
            {
                "anonymous": "never",
                "asyncArrow": "always",
                "constructor": "never",
                "method": "never",
                "named": "never",
            },
        ],
```

rule which does not validate when Eslint is starting:

```
	Configuration for rule "space-before-function-paren" is invalid:
	Value {"anonymous":"never","asyncArrow":"always","constructor":"never","method":"never","named":"never"} should be equal to one of the allowed values.
	Value {"anonymous":"never","asyncArrow":"always","constructor":"never","method":"never","named":"never"} should NOT have additional properties.
	Value {"anonymous":"never","asyncArrow":"always","constructor":"never","method":"never","named":"never"} should match exactly one schema in oneOf.
```

#### Expected Behavior

The rule is correctly generated according to [the spec](https://eslint.org/docs/rules/space-before-function-paren#options). It seems `method` and `constructor` shouldn't be there.

### PR DESCRIPTION
**Pull request by dimabory**
## PR Checklist

-   [x] Addresses an existing issue: fixes #147 

## Overview

- Added filtering of TSLint [rules arguments](https://palantir.github.io/tslint/rules/space-before-function-paren/) to allow only some of them, which are available in [ESLint](https://eslint.org/docs/rules/space-before-function-paren#options).
- Added test.

### DIFF
--- a/src/rules/converters/space-before-function-paren.ts
+++ b/src/rules/converters/space-before-function-paren.ts
@@ -1,11 +1,26 @@
 import { RuleConverter } from "../converter";
 
+const NOT_SUPPORTED_OPTIONS = ["constructor", "method"];
+
+function isExistedESLintOption(rule: string) {
+    return !NOT_SUPPORTED_OPTIONS.includes(rule);
+}
+
 export const convertSpaceBeforeFunctionParen: RuleConverter = tslintRule => {
+    const ruleArguments = tslintRule.ruleArguments.filter(isExistedESLintOption);
+    const notices = NOT_SUPPORTED_OPTIONS.reduce<string[]>((acc, option) => {
+        if (tslintRule.ruleArguments.includes(option)) {
+            acc.push(`Option "${option}" is not supported by ESLint.`);
+        }
+        return acc;
+    }, []);
+
     return {
         rules: [
             {
                 ...(tslintRule.ruleArguments.length !== 0 && {
-                    ruleArguments: tslintRule.ruleArguments,
+                    ruleArguments,
+                    ...(notices.length !== 0 && { notices }),
                 }),
                 ruleName: "space-before-function-paren",
             },
--- a/src/rules/converters/tests/space-before-function-paren.test.ts
+++ b/src/rules/converters/tests/space-before-function-paren.test.ts
@@ -29,4 +29,42 @@ describe(convertSpaceBeforeFunctionParen, () => {
             ],
         });
     });
+
+    test("conversion with all existing arguments", () => {
+        const result = convertSpaceBeforeFunctionParen({
+            ruleArguments: ["anonymous", "named", "asyncArrow", "method", "constructor"],
+        });
+
+        expect(result).toEqual({
+            rules: [
+                {
+                    notices: [
+                        'Option "constructor" is not supported by ESLint.',
+                        'Option "method" is not supported by ESLint.',
+                    ],
+                    ruleArguments: ["anonymous", "named", "asyncArrow"],
+                    ruleName: "space-before-function-paren",
+                },
+            ],
+        });
+    });
+
+    test('conversion with not supported options ["method", "constructor"]', () => {
+        const result = convertSpaceBeforeFunctionParen({
+            ruleArguments: ["method", "constructor"],
+        });
+
+        expect(result).toEqual({
+            rules: [
+                {
+                    notices: [
+                        'Option "constructor" is not supported by ESLint.',
+                        'Option "method" is not supported by ESLint.',
+                    ],
+                    ruleArguments: [],
+                    ruleName: "space-before-function-paren",
+                },
+            ],
+        });
+    });
 });

### CODE REVIEW

File: .gitignore [hunk 1]
```diff
@@ -1,3 +1,4 @@
+.idea
```
JoshuaKGoldberg:
```suggestion
.idea
```

If you're interested in getting this checked in, would you mind sending a separate PR? I don't see any reason not to in general -- it's just good to keep PR changes as small as possible.

src/rules/converters/space-before-function-paren.ts
```diff
@@ -1,11 +1,15 @@
 import { RuleConverter } from "../converter";
 
+function isExistedESLintOption(rule: string) {
+    return ["anonymous", "named", "asyncArrow"].includes(rule);
```
JoshuaKGoldberg: For these three, we'll want to include something in `notices` explaining that they will no longer be checked.

https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/docs/Architecture.md#rule-converters

dimabory: Added notices.

.gitignore
dimabory: Cleaned it up 😏

src/rules/converters/space-before-function-paren.ts
```diff
@@ -1,11 +1,22 @@
...
+}
+
 export const convertSpaceBeforeFunctionParen: RuleConverter = tslintRule => {
+    const tslintRuleArguments = tslintRule.ruleArguments;
```
JoshuaKGoldberg: There's not really much use in making a quick storage variable whose name is only one character away from its stored lookup. So you have any thoughts on removing this?

src/rules/converters/space-before-function-paren.ts
```diff
@@ -1,11 +1,22 @@
...
     return {
         rules: [
             {
-                ...(tslintRule.ruleArguments.length !== 0 && {
-                    ruleArguments: tslintRule.ruleArguments,
+                ...(tslintRuleArguments.length !== 0 && {
+                    ruleArguments,
+                    ...(tslintRuleArguments.length > ruleArguments.length && {
+                        notices: [
+                            'ESLint doesn\'t support the following options: "method" and "constructor". These two options can be covered by "named" therefore make sure you have set this option.',
```
JoshuaKGoldberg: A couple of things:

* It would be a little confusing to have only one of the two unsupported options enabled in TSLint, then get a message about the other one in your log. Maybe this should only mention the option(s) you have enabled?
* I don't follow what you mean by `covered by "named"`. Is that meant to be a helpful hint that the equivalent ESLint rule option to enable is `named`? If so, is the proper output of the converter when both those TSLint options are enabled to have that ESLint option turned on?

src/rules/converters/space-before-function-paren.ts
dimabory: For sure it makes sense.

src/rules/converters/space-before-function-paren.ts
dimabory: The first point is definitely reasonable. I will collect notices properly.

Regarding the second point, I would say I've provided not ideal description `covered by "named"`. _TSLint's_ `"named"` !== _ESLint's_ `"named"`, but `eslint` includes `"constructor"` and `"method"` into the `"named"` option (that's what I mean by `covered by "named"`). Here some examples explaining the difference:

 - tslint [`"space-before-function-paren": [true, { "named": "always" }]`](https://palantir.github.io/tslint-playground/?saved=N4Igxg9gJgpiBcIBmBXAdmALgSwmgBEhBPgBQCU+wAOgfvQE4yYoMGYMowDctAvrVpgANgEMAzuPwAxYlVr16AeiX40JGAwYQGUmNkwALTfkhpxHFFh34bAW2aHo8BYrMXO1hhXrB8SgCoA1QE6egAHFAAjYWwwfAcjaB8-QOD8UNCQABpwPCRsAHMEEBow-GoQGAAPTBg0KHFK+HwAbUrMcVi0THgmSDsHBpgoSoBdbNd6Ss5hGCaEeXLFCpBxcNEwGABaKJgiJm3UDBw8bY2mNGa2yxhsqlW0UQdRxcrRYQB3UQBPBYyxlMMvxaCA+EA) vs. eslint [`space-before-function-paren: [2, { "named" : "always" }]`](https://eslint.org/demo#eyJ0ZXh0IjoiLyplc2xpbnQgc3BhY2UtYmVmb3JlLWZ1bmN0aW9uLXBhcmVuOiBbMiwgeyBcIm5hbWVkXCIgOiBcImFsd2F5c1wiIH1dKi9cbi8qZXNsaW50LWVudiBlczYqL1xuXG5mdW5jdGlvbiBmb28gKCkge31cblxuY2xhc3MgRm9vIHtcbiAgICBjb25zdHJ1Y3RvcigpIHt9XG4gICAgaGVsbG8oKSB7fVxufSIsIm9wdGlvbnMiOnsicGFyc2VyT3B0aW9ucyI6eyJlY21hVmVyc2lvbiI6MTAsInNvdXJjZVR5cGUiOiJtb2R1bGUiLCJlY21hRmVhdHVyZXMiOnt9fSwicnVsZXMiOnt9LCJlbnYiOnt9fX0=)
 - tslint [`"space-before-function-paren": [true, { "constructor": "always", "method": "always" }]`](https://palantir.github.io/tslint-playground/?saved=N4Igxg9gJgpiBcIBmBXAdmALgSwmgBEhBABQCU+wAOgfnQE4yYr0Gb0owDcNAvjTTAAbAIYBnMfgBixSjTp1IaMexRYI9fOTrB8AegBUBvfn606ABxQAjIdjD40IgLYwoWirsPHTfGiAAacDwkbABzBBBqc3wqEBgAD0wYNCgxOPh8AG04zDE7NEx4RkhnV1S3OIBdAPkFOI4hGHSEORiFWJAxCxEwGABaaxgiRn7UDBw8fp7GNAzs1RgAyk6lFQ51enm4kSEAdxEATxbluNdMAAtobZBdg+O40yq6ujMzEF4gA) vs. eslint [`space-before-function-paren: [2, { "named" : "always" }]`](https://eslint.org/demo#eyJ0ZXh0IjoiLyplc2xpbnQgc3BhY2UtYmVmb3JlLWZ1bmN0aW9uLXBhcmVuOiBbMiwgeyBcIm5hbWVkXCIgOiBcImFsd2F5c1wiIH1dKi9cbi8qZXNsaW50LWVudiBlczYqL1xuXG5mdW5jdGlvbiBmb28oKSB7XG4gICAgcmV0dXJuIHRydWU7XG59XG5cbmNsYXNzIEZvbyB7XG4gICAgY29uc3RydWN0b3IgKCkgIHsgLyoqLyB9XG4gICAgbmFtZWQgKCkgeyAvKiovIH1cbn0iLCJvcHRpb25zIjp7InBhcnNlck9wdGlvbnMiOnsiZWNtYVZlcnNpb24iOjEwLCJzb3VyY2VUeXBlIjoibW9kdWxlIiwiZWNtYUZlYXR1cmVzIjp7fX0sInJ1bGVzIjp7fSwiZW52Ijp7fX19)


Do you have ideas on how to make the notice to be simple and understandable at the same time?

JoshuaKGoldberg: What you have now with ` 'Option "constructor" is not supported by ESLint.'` seems nice and simple. I'm pretty mediocre at thinking of these styles, so let's go with that! 🚀

PR #280, opened 2019-12-12
feat: add import-blacklist converter #277

### LINKED ISSUE #277: Missing converter: import-blacklist to no-restricted-imports
### 💥 Missing Converter

-   `tslint-to-eslint-config` version: 0.3.1
-   `typescript-eslint` version: 2.10.0

#### TSLint Rule Name

import-blacklist

#### ESLint Rule Name

no-restricted-imports

#### Example

````
"import-blacklist": [
  true,
  "rxjs/Rx"
],
````
should be converted to:
```
"no-restricted-imports": ["error", {"paths": ["rxjs/Rx"]}],
```

### PR DESCRIPTION
**Pull request by dimabory**
## PR Checklist

-   [x] Addresses an existing issue: fixes #277

## Overview

Adding the new converter [import-blacklist](https://palantir.github.io/tslint/rules/import-blacklist/) --> [no-restricted-imports](https://eslint.org/docs/rules/no-restricted-imports).

### DIFF
--- a/src/rules/converters/import-blacklist.ts
+++ b/src/rules/converters/import-blacklist.ts
@@ -1,0 +1,52 @@
+import { RuleConverter } from "../converter";
+import { RequireAtLeastOne } from "../../utils";
+
+type ESLintOptionPath = {
+    name: string;
+    importNames: string[];
+    message?: string;
+};
+type ESLintSimpleOption = string[];
+type ESLintComplexOption = RequireAtLeastOne<{
+    paths: (string | ESLintOptionPath)[];
+    patterns: string[];
+}>;
+type ESLintOptions = ESLintSimpleOption | ESLintComplexOption;
+
+export const convertImportBlacklist: RuleConverter = tslintRule => {
+    const tslintRules = tslintRule.ruleArguments;
+    let ruleArguments: ESLintOptions = [];
+
+    if (tslintRules.every(isString)) {
+        ruleArguments = tslintRules;
+    } else {
+        ruleArguments = [
+            tslintRules.reduce((rules, rule) => {
+                if (!Array.isArray(rule)) {
+                    const eslintRule = isString(rule)
+                        ? rule
+                        : {
+                              name: Object.keys(rule)[0],
+                              importNames: Object.values(rule)[0] as string[],
+                          };
+                    return { ...rules, paths: [...(rules.paths || []), eslintRule] };
+                } else if (Array.isArray(rule)) {
+                    return { ...rules, patterns: [...(rules.patterns || []), ...rule] };
+                }
+                return rules;
+            }, {} as ESLintComplexOption),
+        ];
+    }
+    return {
+        rules: [
+            {
+                ruleArguments,
+                ruleName: "no-restricted-imports",
+            },
+        ],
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/rules/converters/import-blacklist.ts [hunk 1]
```diff
@@ -0,0 +1,52 @@
...
+                        ? rule
+                        : {
+                              name: Object.keys(rule)[0],
+                              importNames: Object.values(rule)[0] as string[],
+                          };
+                    return { ...rules, paths: [...(rules.paths || []), eslintRule] };
+                } else if (Array.isArray(rule)) {
+                    return { ...rules, patterns: [...(rules.patterns || []), ...rule] };
+                }
+                return rules;
```
JoshuaKGoldberg: Is it.. possible for this case to be hit?

src/rules/converters/import-blacklist.ts
```diff
@@ -0,0 +1,52 @@
...
+                if (!Array.isArray(rule)) {
+                    const eslintRule = isString(rule)
+                        ? rule
+                        : {
+                              name: Object.keys(rule)[0],
+                              importNames: Object.values(rule)[0] as string[],
+                          };
+                    return { ...rules, paths: [...(rules.paths || []), eslintRule] };
+                } else if (Array.isArray(rule)) {
+                    return { ...rules, patterns: [...(rules.patterns || []), ...rule] };
```
JoshuaKGoldberg: ESLint and TSLint use different strategies to match patterns.

* ESLint uses [`ignore`](https://github.com/eslint/eslint/blob/dd1c30e35f05ed332e2abbd3d4d53635efde74b8/lib/rules/no-restricted-imports.js#L11)
* TSLint uses [regular expressions](https://palantir.github.io/tslint/rules/import-blacklist) 

There's nothing we can reasonably do to convert from one form to the other 😢 so the best way to deal with the difference is to add a `notices` entry mentioning the format is different.

src/rules/converters/import-blacklist.ts
dimabory: I'd say no...

src/rules/converters/tests/import-blacklist.test.ts
```diff
@@ -0,0 +1,67 @@
...
+                        {
+                            name: "dummy",
+                            importNames: ["default"],
+                        },
+                    ],
+                    patterns: [".*\\.temp$", ".*\\.tmp$"],
+                },
+            ],
+        ],
+    ] as any[][])("convert %j", (ruleArguments: any[], expected: any[]) => {
```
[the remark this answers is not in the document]
dimabory: Sorry, I didn't get your point regarding this.

----

```
test.each(table)(name, fn, timeout)
Also under the alias: it.each(table)(name, fn) and it.each`table`(name, fn)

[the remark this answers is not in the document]
JoshuaKGoldberg: Ah, apologies, somehow I missed that that was what you were using! Ignore me 😄

dimabory:
> Seems reasonable - amusing type system shenanigans. Just waiting on the one comment about the pattern format. Thanks!

ping

PR #293, opened 2019-12-19
fix `space-before-function-paren`

### LINKED ISSUE #292: Configuration for rule "space-before-function-paren" is invalid
### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 0.4.0
-   ESLint version: 6.7.2
-   Node version: 12.14.0

It was not fixed in #273 .

#### Actual Behavior

`tslint.json`

```
{
  "rules": {
    "space-before-function-paren": [
      true,
      {
        "anonymous": "never",
        "asyncArrow": "always",
        "constructor": "never",
        "method": "never",
        "named": "never"
      }
    ]
  }
}
```

`.eslintrc.js`

```
module.exports = {
    ...
    "rules": {
        "space-before-function-paren": [
            "error",
            {
                "anonymous": "never",
                "asyncArrow": "always",
                "constructor": "never",
                "method": "never",
                "named": "never"
            }
    ...
};
```

`eslint index.ts`

```
Error: .eslintrc.js:
        Configuration for rule "space-before-function-paren" is invalid:
        Value {"anonymous":"never","asyncArrow":"always","constructor":"never","method":"never","named":"never"} should be equal to one of the allowed values.
        Value {"anonymous":"never","asyncArrow":"always","constructor":"never","method":"never","named":"never"} should NOT have additional properties.
        Value {"anonymous":"never","asyncArrow":"always","constructor":"never","method":"never","named":"never"} should match exactly one schema in oneOf.
```

#### Expected Behavior

```
module.exports = {
    ...
    "rules": {
        "space-before-function-paren": [
            "error",
            {
                "anonymous": "never",
                "asyncArrow": "always",
                "named": "never"
            }
    ...
};
```

#### Reproduction

```
#!/bin/bash -x
mkdir example
cd example/
node --version # v12.14.0
npm init -y
npm i -D \
  @typescript-eslint/parser@2.12.0 \
  @typescript-eslint/eslint-plugin@2.12.0 \
  eslint@6.7.2 \
  tslint@5.20.1 \
  typescript@3.7.3
cat <<EOS >tslint.json
{
  "rules": {
    "space-before-function-paren": [
      true,
      {
        "anonymous": "never",
        "asyncArrow": "always",
        "constructor": "never",
        "method": "never",
        "named": "never"
      }
    ]
  }
}
EOS

$(npm bin)/tsc --init

npx tslint-to-eslint-config@0.4.0

cat .eslintrc.js # generated
# module.exports = {
#     "env": {
#         "browser": true,
#         "es6": true
#     },
#     "parser": "@typescript-eslint/parser",
#     "parserOptions": {
#         "project": "tsconfig.json",
#         "sourceType": "module"
#     },
#     "plugins": [
#         "@typescript-eslint"
#     ],
#     "rules": {
#         "space-before-function-paren": [
#             "error",
#             {
#                 "anonymous": "never",
#                 "asyncArrow": "always",
#                 "constructor": "never",
#                 "method": "never",
#                 "named": "never"
#             }
#         ]
#     }
# };

# dummy source file
echo 'function add(x: number, y: number) { return x + y; }' > index.ts

$(npm bin)/eslint index.ts
# Error: .eslintrc.js:
#         Configuration for rule "space-before-function-paren" is invalid:
#         Value {"anonymous":"never","asyncArrow":"always","constructor":"never","method":"never","named":"never"} should be equal to one of the allowed values.
#         Value {"anonymous":"never","asyncArrow":"always","constructor":"never","method":"never","named":"never"} should NOT have additional properties.
#         Value {"anonymous":"never","asyncArrow":"always","constructor":"never","method":"never","named":"never"} should match exactly one schema in oneOf.
```

bouzuya:
## PR Checklist

-   [x] Addresses an existing issue: fixes #292 

## Overview

See: #292


src/rules/converters/space-before-function-paren.ts
```diff
@@ -1,29 +1,32 @@
...
-    return {
-        rules: [
-            {
-                ...(tslintRule.ruleArguments.length !== 0 && {
-                    ruleArguments,
+    const argument: "always" | "never" | { [key: string]: "always" | "never" } | undefined =
+        tslintRule.ruleArguments[0];
+    const ruleName = "space-before-function-paren";
+    if (argument === "always" || argument === "never") {
+        return { rules: [{ ruleArguments: [argument], ruleName }] };
```
JoshuaKGoldberg: I'm going to send a separate change to enable [`no-else-return`](https://eslint.org/docs/rules/no-else-return) in the master branch. Could you please fix its complaints here? That way we don't get surprises when the PR is merged.

src/rules/converters/space-before-function-paren.ts
```diff
@@ -1,29 +1,32 @@
...
+    } else if (typeof argument === "object" && argument !== null) {
+        const notices = Object.keys(argument)
+            .filter(option => !SUPPORTED_OPTIONS.includes(option))
+            .map(option => `Option "${option}" is not supported by ESLint.`);
+        const filtered = Object.keys(argument)
+            .filter(option => SUPPORTED_OPTIONS.includes(option))
+            .reduce((obj, option) => {
+                return { ...obj, [option]: argument[option] };
+            }, {} as { [key: string]: "always" | "never" });
```
JoshuaKGoldberg: Fun fact! This is a little cleaner:

```ts
            .reduce<Record<string, "always" | "never">((obj, option) => {
                    ruleArguments,	                return { ...obj, [option]: argument[option] };
            }, {});
```ts

File: src/rules/converters/space-before-function-paren.ts [hunk 3]
```diff
@@ -1,29 +1,32 @@
...
-        rules: [
-            {
-                ...(tslintRule.ruleArguments.length !== 0 && {
-                    ruleArguments,
+    const argument: "always" | "never" | { [key: string]: "always" | "never" } | undefined =
+        tslintRule.ruleArguments[0];
+    const ruleName = "space-before-function-paren";
+    if (argument === "always" || argument === "never") {
+        return { rules: [{ ruleArguments: [argument], ruleName }] };
+    } else if (typeof argument === "object" && argument !== null) {
```
JoshuaKGoldberg: The contents in the body of this `else if` are a little substantial to read through. Would you mind splitting them into a separate function? That'd be helpful for readability IMO.

src/rules/converters/space-before-function-paren.ts
bouzuya: OK. I will fix it.

src/rules/converters/space-before-function-paren.ts
bouzuya: OK. I will stop using `as`.

src/rules/converters/space-before-function-paren.ts
bouzuya: I'm sorry. I can't read English well. In this what you mean?

```typescript
const isObject = (o: unknown): o is Record<string, "always" | "never"> =>
    typeof o === "object" && o !== null;

// ...
} else if (isObject(argument)) {
// ...
```

Alternatively, I think it would be better to use `typeof argument === "undefined"` if `tslint.json` is valid.

```typescript
// ...
} else if (typeof argument === "undefined") {
    return { rules: [{ ruleName }] };
}
const notices = Object.keys(argument)
    .filter(option => !SUPPORTED_OPTIONS.includes(option))
    .map(option => `Option "${option}" is not supported by ESLint.`);
const filtered = Object.keys(argument)
    .filter(option => SUPPORTED_OPTIONS.includes(option))
    .reduce<Record<string, "always" | "never">>((obj, option) => {
        return { ...obj, [option]: argument[option] };
    }, {});
return {
    rules: [
        {
            ...(notices.length !== 0 && { notices }),
            ...(Object.keys(filtered).length !== 0 && { ruleArguments: [filtered] }),
            ruleName,
        },
    ],
};
```

What would you like to do specifically?

JoshuaKGoldberg: Oh sorry for the ambiguity in my comments! Yes, your second suggestion around using `typeof` seems great. 

My overall goal was to reduce the amount of logic placed inside the `else` statement, since that can be confusing to read through. Moving that stuff into a separate function can be helpful to split the code up so each individual portion is understandable.

bouzuya:
Can I ask to fix them?

JoshuaKGoldberg:
Sure! I'll merge this in now and send a separate PR. Thanks for all your work! 🙌

JoshuaKGoldberg:
Actually, wait, @bouzuya I think I missed a commit - you applied all the changes I asked for. I think I'm just confusing you unnecessarily. Sorry! 😅

PR #316, opened 2020-01-24
[Feature] Add ban rule converter

KingDarBoja:
## PR Checklist

-   [ ] Addresses an existing issue: fixes None

## Overview

Add ban rule converter. Still need to fix the test suite but would like some feedback first.

### DIFF
--- a/src/rules/converters/ban.ts
+++ b/src/rules/converters/ban.ts
@@ -1,0 +1,67 @@
+import { RuleConverter } from "../converter";
+
+type BanRule = {
+    object?: string;
+    property?: string;
+    message?: string;
+};
+
+export const convertBan: RuleConverter = tslintRule => {
+    let bannedObjMethod: BanRule[] = [];
+
+    for (const rule of tslintRule.ruleArguments) {
+        switch (rule.constructor.name) {
+            case "String":
+                bannedObjMethod.push({
+                    property: rule,
+                });
+                break;
+            case "Array":
+                if (rule.length == 1) {
+                    bannedObjMethod.push({
+                        property: rule[0],
+                    });
+                } else {
+                    bannedObjMethod.push({
+                        object: rule[0],
+                        property: rule[1],
+                        ...{ message: rule[2] ? rule[2] : undefined },
+                    });
+                }
+                break;
+            case "Object":
+                if (Array.isArray(rule.name)) {
+                    if (rule.name.length == 2) {
+                        // TODO: Fix iteration if multiple properties.
+                        bannedObjMethod.push({
+                            ...{ object: rule.name[0] !== "*" ? rule.name[0] : undefined },
+                            property: rule.name[1],
+                            ...{ message: rule.message ? rule.message : undefined },
+                        });
+                    } else {
+                        // Don't know how to deal with nested functions.
+                    }
+                } else {
+                    bannedObjMethod.push({
+                        property: rule.name,
+                        ...{ message: rule.message ? rule.message : undefined },
+                    });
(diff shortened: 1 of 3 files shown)


src/rules/converters/ban.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const convertBan: RuleConverter = tslintRule => {
+    let bannedObjMethod: BanRule[] = [];
+
+    for (const rule of tslintRule.ruleArguments) {
+        switch (rule.constructor.name) {
+            case "String":
+                bannedObjMethod.push({
+                    property: rule,
+                });
```
JoshuaKGoldberg: I think this'll need some work. TSLint's `ban` rule bans both _globals_ and _properties_. ESLint splits those between `no-restricted-globals` and `no-restricted-properties`.

For example, this TSLint configuration:

```json
{
    "rules": {
        "ban": [
            true,
            "eval",
        ]
    }
}
```

...is, from my understanding, equivalent to this ESLint configuration:

```json
{
    "rules": {
        "no-restricted-globals": [
            "error",
            "eval"
        ]
    }
}
```

KingDarBoja: I am confused as the roadmap states that it should be only the `no-restricted-properties` rule. [ESLint rule](https://eslint.org/docs/rules/no-restricted-properties)

JoshuaKGoldberg: @KingDarBoja perhaps the roadmap is incorrect?

KingDarBoja: Could be outdated but should be good if someone check that.

KingDarBoja:
Link to the [roadmap.md](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md#functionality) where this rule is listed.

KingDarBoja:
> Link to the [roadmap.md](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md#functionality) where this rule is listed.

@JoshuaKGoldberg should be good if you take a look at this first.

JoshuaKGoldberg:
Asking around for the docs clarification: https://github.com/typescript-eslint/typescript-eslint/pull/1816

KingDarBoja:
> Asking around for the docs clarification: [typescript-eslint/typescript-eslint#1816](https://github.com/typescript-eslint/typescript-eslint/pull/1816)

Is that PR related? It seems like a fix at the readme for another rule rather than a question 🙈

JoshuaKGoldberg:
Oh sure, I thought you meant you'd like to wait for that clarification before continuing. Do you still want to work on this in the interim? If so, bumping the request in the review that output rules should correctly use `no-restricted-globals` and/or `no-restricted-properties`.

KingDarBoja:
@JoshuaKGoldberg I do rather wait for the clarification but once again, it seems you linked the wrong PR/issue 😆

JoshuaKGoldberg:
Today is not my day 🤦‍♂ 😄

JoshuaKGoldberg:
Ok, yes, it was just an oversight. We're good to go!

KingDarBoja:
If that's the case, then it is harder to know which one is a global or a property, unless we rely on something like checking if it's a function or if it's a property 🤔  @JoshuaKGoldberg

KingDarBoja:
I will close this one as I need to rethink how to tackle this rule to properly check when it is a function or a property. In the worst case, just put an notice to let user know that this rule *must* be set manually.

PR #323, opened 2020-02-07
Change notice behavior for trailing-comma esSpecCompliant

### LINKED ISSUE #329: Changing notifice behavior for trailing-comma, esSpecCompliant option
#### Actual Behavior

When `esSpecCompliant` is present, the following notice is given

> ESLint does not support config property esSpecCompliant

#### Expected Behavior

There is however a distinction in the handling of trailing commas in regard to rest params by both linters. Trailing commas after a rest param is a syntax error in modern JavaScript/TypeScript. 

Both rules were implemented without special treatment to rest params. Then ES2015/ES6 came along and introduced them.

TSLint introduced an option (`esSpecCompliant`) to change its rule's behavior to stop requiring commas after rest params. Without this option enabled, a comma is still required despite the syntax error it causes. This was done to not break old TypeScript versions. In all their recommended configs this option is enabled by default. There was an issue that sought to default the option to true but it got closed due to the library's deprecation (https://github.com/palantir/tslint/issues/4172).

ESLint introduced a breaking change and no option can be enabled to require a comma after a rest param (See discussion https://github.com/eslint/eslint/issues/7297).

Currently, tslint-to-eslint-config throws notices for having `esSpecCompliant: true`. I propose to change this because `esSpecCompliant: true` matches the ESLint rule and is the only valid config for modern JS/TS. I want to change the notice behavior so that only `esSpecCompliant: false` gives any notices. Alternatively we could also give notices when `esSpecCompliant` is not specified.

MatthiasKunnen:
## PR Checklist

-   [x] Addresses an existing issue: fixes #329

## Overview

I've changed the notice behavior of `esSpecCompliant` property for rule trailing-comma. Previously, a notice would be given when the `esSpecCompliant` was present. Since eslint only supports `esSpecCompliant == true` no notice should be given in that instance.

After this change, a notice will be given when `esSpecCompliant` is either `undefined` or `false`.

### DIFF
--- a/src/rules/converters/tests/trailing-comma.test.ts
+++ b/src/rules/converters/tests/trailing-comma.test.ts
@@ -22,24 +22,28 @@ describe(convertTrailingComma, () => {
                     singleline: "never",
                 },
                 expectedRuleArguments: [],
+                expectedNotices: ["ESLint only supports esSpecCompliant enabled"],
             },
             {
                 argument: {
                     singleline: "always",
                 },
                 expectedRuleArguments: [],
+                expectedNotices: ["ESLint only supports esSpecCompliant enabled"],
             },
             {
                 argument: {
                     multiline: "never",
                 },
                 expectedRuleArguments: [],
+                expectedNotices: ["ESLint only supports esSpecCompliant enabled"],
             },
             {
                 argument: {
                     multiline: "always",
                 },
                 expectedRuleArguments: ["always-multiline"],
+                expectedNotices: ["ESLint only supports esSpecCompliant enabled"],
             },
             {
                 argument: {
@@ -47,6 +51,7 @@ describe(convertTrailingComma, () => {
                     multiline: "never",
                 },
                 expectedRuleArguments: [],
+                expectedNotices: ["ESLint only supports esSpecCompliant enabled"],
             },
             {
                 argument: {
@@ -54,6 +59,7 @@ describe(convertTrailingComma, () => {
                     multiline: "always",
                 },
                 expectedRuleArguments: ["always-multiline"],
+                expectedNotices: ["ESLint only supports esSpecCompliant enabled"],
             },
             {
                 argument: {
@@ -61,6 +67,7 @@ describe(convertTrailingComma, () => {
                     multiline: "never",
                 },
                 expectedRuleArguments: [],
+                expectedNotices: ["ESLint only supports esSpecCompliant enabled"],
             },
             {
                 argument: {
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
👋 @MatthiasKunnen, thanks for reporting the issue and opening a PR!

I'm a little confused. TSLint's [`trailing-comma` docs](https://palantir.github.io/tslint/rules/trailing-comma) states:

> To align this rule with the ECMAScript specification that is implemented in modern JavaScript VMs, there is a third option `esSpecCompliant`. Set this option to true to **disallow** trailing comma on object and array rest and rest parameters.

...but ESLint's [`comma-dangle` docs](https://eslint.org/docs/rules/comma-dangle) look like they prefer **allowing** the trailing comma. 

Per the PR template, would you mind filing an issue to discuss the changes? In general it'd be preferable to discuss whether changes should happen in issues, and save PRs for just the implementation changes.
_(I can go ahead and do that if you don't have time to soon)_

MatthiasKunnen:
You're absolutely correct about both implementations.

Both TSLint and ESLint implemented their respective rules in a way that did not deal with rest params. Both changed their rules to deal with rest params but they went separate directions. TSLint decided that old versions of typescript should not be broken as they allowed the trailing comma so they added an option. ESLint, however, decided since a trailing comma after a rest param is a syntax error, a breaking change was okay and they since disallow a comma after a rest param (discussion here https://github.com/eslint/eslint/issues/7297).

TSLint actually had a discussion about removing the `esSpecCompliant` option and disallowing the trailing comma after a rest param which would result in the same behavior as ESLint (see https://github.com/palantir/tslint/issues/4172). That issue got closed before any action could be taken due to TSLint's deprecation. The issue mentions the following:

> esSpecCompliant: true is enabled in all the built-in configs, tslint:all, tslint:recommended, and tslint:latest, which I believe should be sufficient for most users. we can change the behavior of the rule to make esSpecCompliant: true the default behavior in v6.0

We can thus assume that `esSpecCompliant` will be true in nearly all configs that have the `trailing-comma` rule. Right now, these configs will receive warnings about ESLint's lack of such a feature. This, however, won't be a problem and for that reason this PR changes the conversion in order to default `esSpecCompliant` to true and warn when it's specifically set to false.

I hope that this explanation is clear as to the purpose of the PR and the underlying reasons. If not I can make an issue.

JoshuaKGoldberg:
> If not I can make an issue.

Yes please!

PR #353, opened 2020-03-30
Respect the original TSLint configuration of `ban-types`

### LINKED ISSUE #352: `ban-types` converter should respect the original TSLint configuration for each type descriptor
### 🚀 Feature Request

Now, `ban-types` converter drops the original TSLint configuration for each type descriptor.
I mean, even if TSLint specified `ban-types` with the type argument, generated ESLint configuration doesn't contain the detailed type configuration.

In this situation, generated configuration only contains `"@typescript-eslint/ban-types": "error"`, but according to the official documentation, probably this generated result
doesn't make sense because each rule requires the type name (and the message).

https://github.com/typescript-eslint/typescript-eslint/blob/f3160b471f8247e157555b6cf5b40a1f6ccdc233/packages/eslint-plugin/docs/rules/ban-types.md

#### Existing Behavior

For example, when the TSLint configuration is like the following:

```
{
  "rules": {
    "ban-types": [true,
      ["Object", "Use {} instead."],
    ]
  }
}
```

then, current logic generates the following ESLint configuration:

```
  "rules": {
      "@typescript-eslint/ban-types": "error"
  }
```

#### Change Proposal

I suppose this should generate the configuration that is like the following:

```
  "rules": {
      "@typescript-eslint/ban-types": [
          "error",
          {
              "types": {
                  "Object": {
                      "message": "Use {} instead."
                  }
              }
          }
      ]
  }
```

### Note

I've already implemented a patch for this issue (but unit-tests are not yet).
https://github.com/typescript-eslint/tslint-to-eslint-config/compare/master...moznion:fix_ban_types?expand=1

Can I send a pull-request for this repository with unit-tests, if you don’t mind?

moznion:
## PR Checklist

-   [x] Addresses an existing issue: fixes #352

## Overview

When it gives the following TSLint configuration:

```
{
  "rules": {
    "ban-types": [true,
      ["Object", "Use {} instead."],
    ]
  }
}
```

as-is:

```
  "rules": {
      "@typescript-eslint/ban-types": "error"
  }
```

maybe this doesn't make sense.

to-be:

```
  "rules": {
      "@typescript-eslint/ban-types": [
          "error",
          {
              "types": {
                  "Object": {
                      "message": "Use {} instead."
                  }
              }
          }
      ]
  }
```

This pull-request enhances a converter for ban-types to make the behaviour to be expected: https://github.com/typescript-eslint/typescript-eslint/blob/f3160b471f8247e157555b6cf5b40a1f6ccdc233/packages/eslint-plugin/docs/rules/ban-types.md

### DIFF
--- a/src/rules/converters/ban-types.ts
+++ b/src/rules/converters/ban-types.ts
@@ -8,12 +8,12 @@ export const convertBanTypes: RuleConverter = tslintRule => {
     const bannedTypes: Record<string, ConvertBanTypeArgument | null> = {};
 
     for (const rule of tslintRule.ruleArguments) {
-        const typ = rule[0];
-        if (!typ) {
+        const bannedType = rule[0];
+        if (!bannedType) {
             break;
         }
 
-        bannedTypes[typ] = rule[1]
+        bannedTypes[bannedType] = rule[1]
             ? {
                   message: rule[1],
               }
--- a/src/rules/converters/ban-types.ts
+++ b/src/rules/converters/ban-types.ts
@@ -1,10 +1,33 @@
 import { RuleConverter } from "../converter";
 
-export const convertBanTypes: RuleConverter = () => {
+export const convertBanTypes: RuleConverter = tslintRule => {
+    type ConvertBanTypeArgument = {
+        message: string;
+    };
+
+    const bannedTypesObj: { [key: string]: ConvertBanTypeArgument | null } = {};
+
+    for (const rule of tslintRule.ruleArguments) {
+        const typ = rule[0];
+        if (!typ) {
+            break;
+        }
+
+        bannedTypesObj[typ] = rule[1]
+            ? {
+                  message: rule[1],
+              }
+            : null;
+    }
+
+    const ruleArguments =
+        Object.keys(bannedTypesObj).length === 0 ? undefined : [{ types: bannedTypesObj }];
+
     return {
         rules: [
             {
                 ruleName: "@typescript-eslint/ban-types",
+                ...{ ruleArguments },
             },
         ],
     };
--- a/src/rules/converters/ban-types.ts
+++ b/src/rules/converters/ban-types.ts
@@ -5,7 +5,7 @@ export const convertBanTypes: RuleConverter = tslintRule => {
         message: string;
     };
 
-    const bannedTypesObj: Record<string, ConvertBanTypeArgument | null> = {};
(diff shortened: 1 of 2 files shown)


src/rules/converters/ban-types.ts
```diff
@@ -1,10 +1,37 @@
 import { RuleConverter } from "../converter";
 
-export const convertBanTypes: RuleConverter = () => {
+export const convertBanTypes: RuleConverter = tslintRule => {
+    type ConvertBanTypeArgument = {
+        message: string;
+    };
+
+    const bannedTypesObj: { [key: string]: ConvertBanTypeArgument | null } = {};
```
JoshuaKGoldberg: Small nit, `Record` is now the built-in type that can handle these kinds of types:

```suggestion
    const bannedTypesObj: Record<string, ConvertBanTypeArgument | null> = {};
```

JoshuaKGoldberg: A few naming nits for this file:

* "`Obj`" is unnecessary in `bannedTypesObj`; could you rename to `bannedTypes`?
* `typ` isn't very clear and I was confused at first as to what it means; could you rename to `bannedType` or similar?

src/rules/converters/ban-types.ts
```diff
@@ -1,10 +1,37 @@
...
+
+    for (const rule of tslintRule.ruleArguments) {
+        if (!Array.isArray(rule)) {
+            break;
+        }
+
+        const typ = rule[0];
```
JoshuaKGoldberg: Nit: if you want to be really slick ✨ 😉 you could do something like:

```ts
const [bannedType, message] = rule;
```

...so you can turn the object assigned to `bannedTypesObj[typ]` into `{ message }`.

JoshuaKGoldberg:
Oh, and feel free to push new commits instead of force pushing. Helps keep the review comments findable 👍

moznion:
Thanks for your lightning-fast reviews! I'll tweak them!
(and I'm sorry to do force-push, the motivation was keeping clean this pull request...)

moznion:
@JoshuaKGoldberg I've just pushed tweaked commits. Could you please review them again?

moznion:
Oops, somehow I removed logic that checks a rule is array object or not. I added that at 3d8ded3cf9dfd01f32185e80925d373a4fdd6d26.

JoshuaKGoldberg:
Thanks @moznion!

moznion:
Thank you for your support, @JoshuaKGoldberg!

PR #318, opened 2020-01-29
always set markers for spaced-comment rule to support reference types

beenotung:
## PR Checklist

-   [x] Addresses an existing issue: fixes #317
-   [x] That issue was marked as [`status: accepting prs`](https://github.com/typescript-eslint/tslint-to-eslint-config/labels/status%3A%20accepting%20prs)

## Overview

Always set markers for spaced-comment rule to support reference types.

Example:
```ts
/// <reference types="path-to-file" />
```

PR #386, opened 2020-04-14
Adjusted comment-format's check-space output for markers

### LINKED ISSUE #317: Conflict on reference types
### 🚀 Feature Request

As mentioned in the issue:
https://github.com/typescript-eslint/typescript-eslint/issues/600

The eslint rule `spaced-comment` does not support reference types syntax, example:
```
/// <reference types="./global" />
```
Will be converted into:
```
// / <reference types="./global" />
```

A configuration in the eslint can fix this:
```
/* eslint spaced-comment: ["error", "always", { "markers": ["/"] }] */
```

However this config is not set by default when converted from `tslint-to-eslint-config`

#### Existing Behavior

Below rule is generated during conversion:
```
        "spaced-comment": "error",
```

#### Change Proposal

Below rule should be generated during conversion:
```
        "spaced-comment": [
            "error",
            "always",
            {
                "markers": ["/"]
            }
        ],
```

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #317

## Overview

Continues #318, but I think corrects the behavior of the rule. TSLint doesn't enforce spacing behavior unless `check-space` is provided, so I think this converter actually should only output a `spaced-comment` configuration when TSLint's `check-space` argument is provided, right? Thoughts @beenotung @KingDarBoja? 

https://palantir.github.io/tslint/rules/comment-format
https://eslint.org/docs/rules/spaced-comment

### DIFF
--- a/src/rules/converters/comment-format.ts
+++ b/src/rules/converters/comment-format.ts
@@ -7,14 +7,10 @@ type CommentFormatOptions = {
 
 export const CapitalizedIgnoreMessage = "Only accepts a single string pattern to be ignored.";
 
-export const convertCommentFormat: RuleConverter = tslintRule => {
+export const convertCommentFormat: RuleConverter = (tslintRule) => {
     const capitalizedRuleArguments: string[] = [];
-    const spaceCommentRuleArguments: string[] = [];
     const capitalizedNotices: string[] = [];
-
-    if (!tslintRule.ruleArguments.includes("check-space")) {
-        spaceCommentRuleArguments.push("never");
-    }
+    const checkSpace = tslintRule.ruleArguments.includes("check-space");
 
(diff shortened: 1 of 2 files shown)


src/rules/converters/comment-format.ts
```diff
@@ -7,14 +7,10 @@ type CommentFormatOptions = {
 
 export const CapitalizedIgnoreMessage = "Only accepts a single string pattern to be ignored.";
 
-export const convertCommentFormat: RuleConverter = tslintRule => {
+export const convertCommentFormat: RuleConverter = (tslintRule) => {
```
JoshuaKGoldberg, line 10:
```suggestion
export const convertCommentFormat: RuleConverter = tslintRule => {
```

PR #413, opened 2020-04-26
Add one-line rule converter

### LINKED ISSUE #410: Missing converter: one-line
### 💥 Missing Converter

-   `tslint-to-eslint-config` version: 0.6.0
-   `typescript-eslint` version: 2.29.0

#### TSLint Rule Name

`one-line`

See https://github.com/typescript-eslint/tslint-to-eslint-config/issues/401 for a reproduction. 

The [typescript-eslint roadmap](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md) indicates `brace-style` is probably the ESLint rule to convert to.

KingDarBoja:
## PR Checklist

-   [x] Addresses an existing issue: fixes #410

## Overview

Add `one-line` TSLint converter to ESLint `brace-style` which will check all kind of tokens using the _one true brace style_ option.

### DIFF
--- a/src/rules/converters/one-line.ts
+++ b/src/rules/converters/one-line.ts
@@ -1,0 +1,17 @@
+import { RuleConverter } from "../converter";
+
+export const CheckAllTokensMsg = "ESLint's brace-style will check all tokens.";
+
+export const convertOneLine: RuleConverter = (tslintRule) => {
+    return {
+        rules: [
+            {
+                ...(tslintRule.ruleArguments.length !== 0 && {
+                    notices: [CheckAllTokensMsg],
+                }),
+                ruleArguments: ["1tbs"],
+                ruleName: "brace-style",
+            },
+        ],
+    };
+};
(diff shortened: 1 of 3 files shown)


src/rules/converters/one-line.ts
```diff
@@ -0,0 +1,17 @@
+import { RuleConverter } from "../converter";
+
+export const CheckAllTokensMsg = "ESLint's brace-style will check all tokens.";
+
+export const convertOneLine: RuleConverter = (tslintRule) => {
+    return {
+        rules: [
+            {
+                ...(tslintRule.ruleArguments.length !== 0 && {
```
JoshuaKGoldberg, line 9: Is this logic backwards? From my understanding of https://palantir.github.io/tslint/rules/one-line and https://github.com/typescript-eslint/tslint-to-eslint-config/issues/410#issuecomment-619482577, TSLint will only check syntax constructs when enabled as a rule argument. 

Assuming that's true, there should be three cases here:

* If `one-line` has _no_ optional params enabled, it's effectively doing nothing, and can be ignored
* If `one-line` has _some_ optional params enabled, the `notice` should mention the other syntax constructs will now be linted
* If `one-line` has _all_ optional params enabled, there shouldn't be a notice*
    * *...unless there are syntax constructs ESLint catches that TSLint didn't? In which case there should be a notice for them.

KingDarBoja: The logic seems to make sense as the current TSLint description made me think the rule was always active and the options were used to **only** check the specified tokens.

Therefore, I made the changes while offline, right now had no internet at all 😆

PR #424, opened 2020-04-27
Removed already-installed packages from install command

### LINKED ISSUE #416: Don't request installing packages already present in package.json
### 🚀 Feature Request

Right now, `tslint-to-eslint-config` will report any packages that must be required for the generated ESLint configuration:

We already read from `package.json` and support a `--package` flag. We could save a little bit of CLI output space skipping printing packages that are already listed in `package.json` as a `dependency` or `devDependency`.

#### Existing Behavior

```
⚡ 7 packages are required for this ESLint configuration. ⚡
  npm install @typescript-eslint/eslint-plugin @typescript-eslint/parser @typescript-eslint/eslint-plugin-tslint eslint eslint-plugin-import eslint-plugin-jsdoc eslint-plugin-prefer-arrow --save-dev
```

#### Change Proposal

If all those packages already exist, the :zap: part of CLI output could be skipped.

If only some already exist, we could skip printing them.

```
⚡ 3 new packages are required for this ESLint configuration. ⚡
  npm install @typescript-eslint/eslint-plugin-tslint eslint-plugin-jsdoc eslint-plugin-prefer-arrow --save-dev
```

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #416

## Overview

Separates `logMissingPackages` out from the multi-purpose `reportConversionResults` and directly calls it in `convertConfig`. It'll now take in the result of reading `--package` (normally `package.json`) and remove any package names contained in `dependencies` or `devDependencies` from the list of suggested packages to install.

Screenshot of the trim new CLI output

Per the notes in #416, we can't look at the `extends` value of an ESLint configuration file for packages that need to be installed - even though `eslint-config-prettier` isn't added to `plugins`. This PR adds in a special case for `eslint-config-prettier` existing in an `extends`. I don't know of any other ones we'll need to add to that special case... for now.

Also adds `plugin:` before plugin names, because that seems to be recommended by the ESLint docs.


src/input/findPackagesConfiguration.ts
```diff
@@ -32,11 +23,9 @@ export const findPackagesConfiguration = async (
         : {
               dependencies: {
                   ...rawConfiguration.dependencies,
-                  ...defaultPackagesConfiguration.dependencies,
               },
               devDependencies: {
                   ...rawConfiguration.devDependencies,
-                  ...defaultPackagesConfiguration.devDependencies,
```
JoshuaKGoldberg, line 39: I don't remember why this was added before, but it isn't doing anything... 🤷

Issue #421, opened 2020-04-26
Add issues for converters for Codelyzer TSLint rules


JoshuaKGoldberg:
### 🚀 Feature Request

[Codelyzer](https://github.com/mgechev/codelyzer) provides a set of great TSLint rules around the Angular area. tslint-to-eslint-config should add converters for those rules.

There's a rules list here: https://github.com/mgechev/codelyzer#rules-status. There are quite a lot of them, so for now let's use this issue to track all of them.

FYI @mgechev

Edit: Oh, and see https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/docs/Architecture.md#rule-converters for a description of how the rule converters work. 🔧

KingDarBoja:
This is a nice addition indeed but should be good to link the output ESLint rules as those are at [angular-eslint](https://github.com/angular-eslint/angular-eslint) repo.

JoshuaKGoldberg:
Looks like these are pretty smooth to convert so far. I'll file separate issues for each of the Codelyzr rules and put them in the [Angular Support](https://github.com/typescript-eslint/tslint-to-eslint-config/projects/1) project. I'll consider this issue the tracking one for doing that work.

KingDarBoja:
@JoshuaKGoldberg Sounds good to me, I will work on more of them on the weekend, expect several PRs 🚀

JoshuaKGoldberg:
These are all filed and tracked in https://github.com/typescript-eslint/tslint-to-eslint-config/projects/1.

JoshuaKGoldberg:
Just following up: all rules now are either marked as Blocked or have a corresponding PR. Cool

highvolt-dev:
Found a number of rules had been implemented in `angular-eslint` - 8 that my project had missing conversions for.

https://github.com/typescript-eslint/tslint-to-eslint-config/issues/494#issuecomment-787105536

highvolt-dev:
Opened PRs for:

- `template-accessibility-alt-text` (#968)
- `template-accessibility-elements-content` (#969)
- `template-accessibility-label-for` (#973)
- `template-accessibility-table-scope` (#972)
- `template-accessibility-valid-aria` (#971)
- `template-click-events-have-key-events` (#970)
- `template-mouse-events-have-key-events` (#974)
- `template-no-distracting-elements` (#975)

JoshuaKGoldberg:
[...]

PR #432, opened 2020-04-29
Add codelyzer component-class-suffix converter

KingDarBoja:
## PR Checklist

-   [x] Addresses an existing issue: references #421

## Overview

This is more like a example of how **codelyzer** rules should be placed inside their own folder so any contributor knows where find those converters.

This rule is pretty simple and examples can be found on [angular-eslint/component-class-suffix test](https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin/tests/rules/component-class-suffix.test.ts) file.

The [TSLint rule](http://codelyzer.com/rules/component-class-suffix/) is pretty much the same as the ESLint one, a set of rule arguments will be the suffix to match for any class decorated with `@Component`.

### DIFF
--- a/src/rules/converters/codelyzer/component-class-suffix.ts
+++ b/src/rules/converters/codelyzer/component-class-suffix.ts
@@ -5,7 +5,11 @@ export const convertComponentClassSuffix: RuleConverter = (tslintRule) => {
         rules: [
             {
                 ...(tslintRule.ruleArguments.length !== 0 && {
-                    ruleArguments: tslintRule.ruleArguments,
+                    ruleArguments: [
+                        {
+                            suffixes: tslintRule.ruleArguments,
+                        },
+                    ],
                 }),
                 ruleName: "@angular-eslint/component-class-suffix",
             },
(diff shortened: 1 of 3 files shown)


src/rules/converters/codelyzer/component-class-suffix.ts
```diff
@@ -0,0 +1,15 @@
+import { RuleConverter } from "../../converter";
+
+export const convertComponentClassSuffix: RuleConverter = (tslintRule) => {
+    return {
+        rules: [
+            {
+                ...(tslintRule.ruleArguments.length !== 0 && {
+                    ruleArguments: tslintRule.ruleArguments,
```
JoshuaKGoldberg, line 8: The [options](https://github.com/angular-eslint/angular-eslint/blob/e3a5796f28b20a6b6e64ba752f5a2d351b778109/packages/eslint-plugin/tests/rules/component-class-suffix.test.ts#L55) in angular-eslint look like they use an object with a `suffixes` member.

KingDarBoja: Done! @JoshuaKGoldberg you're up to review again :)

PR #577, opened 2020-05-11
Added Codelyzer template-cyclomatic-complexity converter

### LINKED ISSUE #503: Missing Codelyzr (Angular) rule: template-cyclomatic-complexity
### Missing Converter

#### TSLint Rule Name

template-cyclomatic-complexity

#### Package Source

**Codelyzr**: http://codelyzer.com/rules/template-cyclomatic-complexity

Splitting out from #421, this issue tracks adding in a missing Codelyzr rule converter.

bbarry:
## PR Checklist

-   [x] Addresses an existing issue: fixes #503

## Overview

I don't think this is quite correct because it seems sometimes you will need to put these rules in a different section based on what style of templates you are using...

If you have your templates in separate html files you need an eslint config like this:

```
{
  "rules": { ... },
  "overrides": [
    {
      "files": ["*.component.html"],
      "parser": "@angular-eslint/template-parser",
      "plugins": ["@angular-eslint/template"],
      "rules": {
        "@angular-eslint/template/cyclomatic-complexity": "error"
      }
    }
  ]
}
```

but if you have inline templates you need:

```
{
  "rules": { ... },
  "overrides": [
    {
      "files": ["*.component.ts"],
      "parser": "@angular-eslint/template-parser",
      "plugins": ["@angular-eslint/template"],
      "rules": {
        "@angular-eslint/template/cyclomatic-complexity": "error"
      }
    }
  ]
}
```

Maybe I am misunderstanding how angular-eslint is supposed to work though...


src/rules/converters/codelyzer/template-cyclomatic-complexity.ts
```diff
@@ -0,0 +1,12 @@
+import { RuleConverter } from "../../converter";
+
+export const convertTemplateCyclomaticComplexity: RuleConverter = () => {
+    return {
+        rules: [
+            {
+                ruleName: "@angular-eslint/template/cyclomatic-complexity",
+            },
```
KingDarBoja, line 8: This rulename should map the input rule arguments into the `maxComplexity` object property. Take a look at [component-class-suffix](https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/src/rules/converters/codelyzer/component-class-suffix.ts) rule converter to see how it is done.

KingDarBoja: Also, I am not sure if the `ruleName` has the proper value as it seems nested inside `eslint-plugin-template` directory. [ESLint cyclomatic-complexity](https://github.com/angular-eslint/angular-eslint/blob/master/packages/eslint-plugin-template/src/rules/cyclomatic-complexity.ts)

bbarry: Changes made; tslint rule has example `[true, 4]` angular-eslint expects `{ maxComplexity: 4 }` and it appears `tslintRule.ruleArguments` skips the `true`.

PR #601, opened 2020-06-01
improve type safety with regards to  config files

bbarry:
## PR Checklist

-   [ ] Addresses an existing issue: fixes #000

## Overview

Last week I said I would look deeper at the poor `DeepPartial<T>` type in https://github.com/typescript-eslint/tslint-to-eslint-config/pull/589

It seems that this type needs to touch a number of config related places and there are several places where incomplete configurations might have slipped through but were fine according to tsc.

### DIFF
--- a/src/input/findReportedConfiguration.test.ts
+++ b/src/input/findReportedConfiguration.test.ts
@@ -1,5 +1,5 @@
 import { createStubExec, createStubThrowingExec } from "../adapters/exec.stubs";
-import { findReportedConfiguration, DeepPartial } from "./findReportedConfiguration";
+import { findReportedConfiguration } from "./findReportedConfiguration";
 
 describe("findReportedConfiguration", () => {
     it("returns stderr as an error when the command fails with a zero exit code", async () => {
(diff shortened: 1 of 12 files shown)


src/input/findReportedConfiguration.test.ts
```diff
@@ -56,4 +56,23 @@ describe("findReportedConfiguration", () => {
             rules,
         });
     });
+
+    it("declares a correct DeepPartial type", () => {
```
JoshuaKGoldberg, line 60: Hmm, I don't think we particularly need to unit test for type system types. I think now that you've fleshed out usage of the `DeepPartial` type, it's apparent through the rest of the code whether and how it works.

What do you think?

src/rules/convertRules.test.ts
```diff
@@ -5,6 +5,19 @@ import { RuleConverter, ConversionResult } from "./converter";
 import { RuleMerger } from "./merger";
 
 describe("convertRules", () => {
+    it("doesn't crash with a type error when passed undefined configuration", () => {
```
JoshuaKGoldberg, line 8: Just nitpicking here: technically a type error wouldn't do anything at runtime.

```suggestion
    it("doesn't crash when passed an undefined configuration", () => {
```

src/input/findReportedConfiguration.test.ts
bbarry: Fine by me, the test is a little goofy compared to other tests in this project. I just wrote it first and then fixed all the rest of the code by iteratively rebuilding and running tests and making small changes until everything worked out.

I'll drop the test in another commit shortly.

src/rules/convertRules.test.ts
bbarry: image

JoshuaKGoldberg: Oh haha good point. You're right. 

I think the gist of the suggestion is correct for a different reason though: in general, other unit tests don't mention the particular error type they might crash with, just that the action succeeds. The naming here confused me.

This is definitely a small thing and non-blocking. I'd be happy to merge if you disagree and do nothing. 😄

bbarry: I'll make the change; the test case is more general than my particular text suggests.

PR #617, opened 2020-06-22
fix: converters now uses naming-convention

### LINKED ISSUE #596: class-name convert deprecated. Rename to naming-convention
It's ot actually missing, but rule name is out of date.
[typescript-eslint issue with explaination](https://github.com/typescript-eslint/typescript-eslint/issues/1854#issuecomment-609913921)
I've a PR, so you can let me know when you accept it and I'll submit.

### 💥 Missing Converter

-   `tslint-to-eslint-config` version: latest
-   `typescript-eslint` version: 3.0.2

#### TSLint Rule Name

**class-name**

KingDarBoja:
## PR Checklist

-   [x] Addresses an existing issue: fixes #596

## Overview

This should bring the latest changes from typescript-eslint (the rule removals section) as requested but not sure if I have to provide a merger for these converters @JoshuaKGoldberg

### DIFF
--- a/src/rules/converters/class-name.ts
+++ b/src/rules/converters/class-name.ts
@@ -4,7 +4,7 @@ export const convertClassName: RuleConverter = () => {
     return {
         rules: [
             {
-                ruleName: "@typescript-eslint/class-name-casing",
+                ruleName: "@typescript-eslint/naming-convention",
             },
         ],
     };
--- a/src/rules/converters/interface-name.ts
+++ b/src/rules/converters/interface-name.ts
@@ -1,10 +1,23 @@
 import { RuleConverter } from "../converter";
 
-export const convertInterfaceName: RuleConverter = () => {
+export const convertInterfaceName: RuleConverter = (tslintRule) => {
     return {
         rules: [
             {
                 ruleName: "@typescript-eslint/naming-convention",
+                ...(tslintRule.ruleArguments.length !== 0 && {
+                    rules: [
+                        {
+                            selector: "interface",
+                            format: ["PascalCase"],
+                            custom: {
+                                regex: "^I[A-Z]",
+                                match:
+                                    tslintRule.ruleArguments[0] === "always-prefix" ? true : false,
+                            },
+                        },
+                    ],
+                }),
             },
         ],
     };
--- a/src/rules/converters/interface-name.ts
+++ b/src/rules/converters/interface-name.ts
@@ -4,7 +4,7 @@ export const convertInterfaceName: RuleConverter = () => {
     return {
         rules: [
             {
-                ruleName: "@typescript-eslint/interface-name-prefix",
+                ruleName: "@typescript-eslint/naming-convention",
             },
         ],
     };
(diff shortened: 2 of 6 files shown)


KingDarBoja:
Also related to this issue at [typescrip-eslint](https://github.com/typescript-eslint/typescript-eslint/issues/2251).

src/rules/converters/class-name.ts
```diff
@@ -4,7 +4,7 @@ export const convertClassName: RuleConverter = () => {
     return {
         rules: [
             {
-                ruleName: "@typescript-eslint/class-name-casing",
+                ruleName: "@typescript-eslint/naming-convention",
```
JoshuaKGoldberg, line 7: [TSLint's `class-name`](https://palantir.github.io/tslint/rules/class-name) only allows PascalCase class names. Reading through the `naming-convention` source, I'm not seeing anything that defaults class names to any particular case: PascalCase or otherwise. That means this'll need some rule options to replicate the old behavior, right?

src/rules/converters/interface-name.ts
```diff
@@ -4,7 +4,7 @@ export const convertInterfaceName: RuleConverter = () => {
     return {
         rules: [
             {
-                ruleName: "@typescript-eslint/interface-name-prefix",
+                ruleName: "@typescript-eslint/naming-convention",
```
JoshuaKGoldberg, line 7: Similar to `class-name`, [TSLint's `interface-name`](https://palantir.github.io/tslint/rules/interface-name/) specifies a default behavior -- and I'm thinking this converter is actually already buggy (at least per docs, if not in reality) because `interface-name`'s docs says it requires one of either `"always-prefix"` or "`never-prefix`".

src/rules/converters/class-name.ts
KingDarBoja: Based on the docs, the [`default` selector](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md#group-selectors) matches everything but it is marked as camelCase, so you got a point here.

src/rules/converters/interface-name.ts
KingDarBoja: Same as above comment, I will take a look this weekend :)

src/rules/converters/class-name.ts
KingDarBoja: @JoshuaKGoldberg Actually, the PascalCase for Classes and Interfaces is set as default. Checkout https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/rules/naming-convention.ts#L331 and the docs at [options](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md#options) (scroll down a bit).

src/rules/converters/interface-name.ts
KingDarBoja: This one is easy, there is an example at the end of the docs: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md#enforce-that-interface-names-do-not-begin-with-an-i

JoshuaKGoldberg: Well, sweet.

PR #638, opened 2020-07-18
Added jsx-key converter

### LINKED ISSUE #521: Missing tslint-react rule: jsx-key
### Missing Converter

#### TSLint Rule Name

jsx-key

#### Package Source

**tslint-react**: https://github.com/palantir/tslint-react (see [#rules](https://github.com/palantir/tslint-react#rules))

Equivalents for tslint-react rules generally live in [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react). If an ESLint rule equivalent doesn't yet exist, this tslint-to-eslint-config issue can be closed until that equivalent is created.

KingDarBoja:
## PR Checklist

-   [x] Addresses an existing issue: fixes #521

## Overview

Based on [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md), provides the correct converter for this rule.

However, I am not 100% sure about the `<enabled>` rule argument as this doesn't seem to be a string but TS code will mark it as a property if it isn't.

### DIFF
--- a/src/rules/converters/eslint-plugin-react/jsx-key.ts
+++ b/src/rules/converters/eslint-plugin-react/jsx-key.ts
@@ -5,7 +5,6 @@ export const convertJsxKey: RuleConverter = () => {
         rules: [
             {
                 ruleName: "react/jsx-key",
-                ruleArguments: ["<enabled>"],
             },
         ],
         plugins: ["eslint-plugin-react"],
--- a/src/rules/converters/eslint-plugin-react/jsx-key.ts
+++ b/src/rules/converters/eslint-plugin-react/jsx-key.ts
@@ -1,0 +1,13 @@
+import { RuleConverter } from "../../converter";
+
+export const convertJsxKey: RuleConverter = () => {
+    return {
+        rules: [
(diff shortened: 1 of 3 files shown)


src/rules/converters/eslint-plugin-react/jsx-key.ts
```diff
@@ -0,0 +1,13 @@
+import { RuleConverter } from "../../converter";
+
+export const convertJsxKey: RuleConverter = () => {
+    return {
+        rules: [
+            {
+                ruleName: "react/jsx-key",
+                ruleArguments: ["<enabled>"],
```
JoshuaKGoldberg, line 8: Yeah, looking through the [source](https://github.com/palantir/tslint-react/blob/b92fb075375394f84594853b928e670d729a1ed8/src/rules/jsxKeyRule.ts#L59) for the TSLint side, they don't account for JSX fragments either. So I guess the rule should have no rule arguments either, until the next major version of `eslsint-plugin-react`.

PR #657, opened 2020-07-29
Added jsx-ban-elements converter

### LINKED ISSUE #516: Missing tslint-react rule: jsx-ban-elements
### Missing Converter

#### TSLint Rule Name

jsx-ban-elements

#### Package Source

**tslint-react**: https://github.com/palantir/tslint-react (see [#rules](https://github.com/palantir/tslint-react#rules))

Equivalents for tslint-react rules generally live in [eslint-plugin-react](https://github.com/yannickcr/eslint-plugin-react). If an ESLint rule equivalent doesn't yet exist, this tslint-to-eslint-config issue can be closed until that equivalent is created.

himgupta72:
## PR Checklist

-   [x] Addresses an existing issue: fixes #516 

## Overview

Adding convertor for jsx-ban-elements.

[sample]
**tslint rule:**
```
_"jsx-ban-elements": {
      "options": [
          ["span", "Use div instead."],
          ["button","Use button instead"],
          ["div"]
      ]
    },_
```

**to Eslint rule:** 

```
_"react/forbid-elements": [
            "error",
            {
                "forbid": [
                    {
                        "element": "span",
                        "message": "Use div instead."
                    },
                    {
                        "element": "button",
                        "message": "Use button instead"
                    },
                    {
                        "element": "div"
                    }
                ]
            }
        ],_
```

**EDIT** See https://github.com/palantir/tslint-react/blob/0ebaba30ac199643c43f5eb567ba65c038bfd8b4/src/rules/jsxBanElementsRule.ts to compare options.

### DIFF
--- a/src/rules/converters/eslint-plugin-react/jsx-ban-elements.ts
+++ b/src/rules/converters/eslint-plugin-react/jsx-ban-elements.ts
@@ -17,21 +17,17 @@ const collectArguments = (ruleArguments: any[]) => {
         return undefined;
     }
 
-    const forbidArr: any[] = [];
-    for (const element of ruleArguments) {
-        const forbidObj: any = {};
-        forbidObj.element = element[0];
-
-        if (element.length == 2) {
-            forbidObj.message = element[1];
-        }
-        forbidArr.push(forbidObj);
-    }
+    const forbiddenElements = ruleArguments.map((ruleArgument) => {
+        return {
+            element: ruleArgument[0],
+            ...(ruleArgument.length === 2 && { message: ruleArgument[1] }),
+        };
+    });
 
     return {
         ruleArguments: [
             {
-                forbid: forbidArr,
+                forbid: forbiddenElements,
             },
         ],
     };
--- a/src/rules/converters/eslint-plugin-react/jsx-ban-elements.ts
+++ b/src/rules/converters/eslint-plugin-react/jsx-ban-elements.ts
@@ -17,16 +17,17 @@ const collectArguments = (ruleArguments: any[]) => {
         return undefined;
     }
 
-    let forbidArr: any[] = [];
-    for (let i = 0; i < ruleArguments.length; i++) {
-        let forbidObj: any = {};
-        forbidObj.element = ruleArguments[i][0];
+    const forbidArr: any[] = [];
+    for (const element of ruleArguments) {
+        const forbidObj: any = {};
+        forbidObj.element = element[0];
 
-        if (ruleArguments[i].length == 2) {
-            forbidObj.message = ruleArguments[i][1];
+        if (element.length == 2) {
+            forbidObj.message = element[1];
         }
         forbidArr.push(forbidObj);
     }
+
     return {
         ruleArguments: [
             {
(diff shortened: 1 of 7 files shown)


src/rules/converters/eslint-plugin-react/jsx-ban-elements.ts
```diff
@@ -0,0 +1,38 @@
...
+        plugins: ["eslint-plugin-react"],
+    };
+};
+
+const collectArguments = (ruleArguments: any[]) => {
+    if (ruleArguments.length === 0) {
+        return undefined;
+    }
+
+    const forbidArr: any[] = [];
```
KingDarBoja, line 20: I think it should be a string type since you can have several items being the regex to match as the first item and the second one is the optional explanation as stated at https://github.com/palantir/tslint-react/blob/0ebaba30ac199643c43f5eb567ba65c038bfd8b4/src/rules/jsxBanElementsRule.ts#L26

himgupta72: Thanks @KingDarBoja for reviewing. Yes, i agree 2nd item is optional here, thats why i have added the length check. 
I am making adding custom object based on the the input length in forbidArr, so i am not sure why it should be string type. Let me know if i am missing anything.

KingDarBoja: It is explained on the above link:

```
options: {
            type: "list",
            listType: {
                type: "array",
                items: { type: "string" },
                minLength: 1,
                maxLength: 2,
            },
        },
```

Also there is an example:

`optionExamples: [[true, ["Object", "Use {} instead."], ["String"]]],`

Based on that, seems to be an array of array of strings :)

JoshuaKGoldberg: Indeed, a `[string, string][]` would be a more accurate type definition.

src/rules/converters/eslint-plugin-react/jsx-ban-elements.ts
```diff
@@ -0,0 +1,38 @@
...
+    for (const element of ruleArguments) {
```
JoshuaKGoldberg, line 21: Mild style nitpicking here: this `for` loop has the sole purpose of pushing to an array, so I think it'd be a little cleaner to use a `.map` instead.

```ts
const forbiddenElements = ruleArguments.map((ruleArgument) => {
    return {
        element: element[0],
        ...(element.length === 2 && { message: element[1] })
    };
});
```

_(also, `===` instead of `==` please)_

src/rules/converters/eslint-plugin-react/jsx-ban-elements.ts
JoshuaKGoldberg: Here's an example config from the source project: https://github.com/palantir/tslint-react/blob/0ebaba30ac199643c43f5eb567ba65c038bfd8b4/src/rules/jsxBanElementsRule.ts

JoshuaKGoldberg (approved):
Looks great, thanks for bearing with us @himgupta72! 🙌

PR #672, opened 2020-08-08
Used commander's .opts() to parse raw argv

### LINKED ISSUE #660: CLI Flags do not work for nodejs 14
### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 1.1.0
-   ESLint version: 7.5.0
-   Node version: 14.6.0

#### Actual Behavior

I just used the recommended command `npx tslint-to-eslint-config --prettier` to convert the lint config. However, I always get the suggestion in the logging message: 
```
☠ Prettier plugins are missing from your configuration. ☠
  We highly recommend running tslint-to-eslint-config --prettier to disable formatting ESLint rules.
  See https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/docs/FAQs.md#should-i-use-prettier.
```
Meanwhile, no configurations about prettier would be inserted in the generted .eslintrc.js file.

I wonder what's wrong here. Did I forget some required plugins or  install wrong version of those dependencies? 
I know some more dependencies are requied here for further use together with eslint. I just want to first make sure the convertion of lint configuration is correct. Unfortunately, it doesn't work for me. 

#### Expected Behavior

No more suggestions in the console message. 
Relevant config values for prettier would be added into the .eslintrc.js

#### Reproduction

Use nodejs 14.6.0

```
npm install -g @angular/cli
ng new my-dream-app
cd my-dream-app
npm i --save-dev eslint
npx tslint-to-eslint-config --prettier
```

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #660

## Overview

As of either the latest Node version or latest commander version (I haven't checked but suspect commander), member variables such as `.prettier` aren't being set on the `command` object. They're now available on `.opts()`.

### DIFF
--- a/src/cli/runCli.ts
+++ b/src/cli/runCli.ts
@@ -35,7 +35,7 @@ export const runCli = async (
 
     const parsedArgv = {
         config: "./.eslintrc.js",
-        ...(command.parse(rawArgv) as Partial<TSLintToESLintSettings>),
+        ...command.parse(rawArgv).opts(),
     };
 
     const programOptions = command.opts();


KingDarBoja:
Yeah, I updated to latest version since renovate bot bumped it and had issues due to breaking changes. Didn't expect the parse method to require the `opts` call too D:

PR #681, opened 2020-08-14
Add notice to arrow-return-shorthand converter

### LINKED ISSUE #345: arrow-return-shorthand isn't completely compatible with arrow-body-style
### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 0.5.1
-   ESLint version: 6.8.0
-   Node version: 8.17.0

#### Actual Behavior

`arrow-return-shorthand` is converted to `arrow-body-style` automagically without any warning.

#### Expected Behavior

Conversion should produce a message about an incompatible use case - when arrow function body is multiline and using curly brackes, but has only one line which returns the result immediately, `arrow-return-shorthand` doesn't yield an error (unless `"multiline"` is specified), while `arrow-body-style` produces an error, suggesting that the result could be returned using the short notation.

I agree that this use case is rather arguable, but in my organization, there's a huge monorepo where such use case is a pattern which is already in place, and it took me a while to figure out what exactly caused the mismatch. Some warning message instead would've been nice :slightly_smiling_face:

In my case, I actually commented this rule out, until further development of things (either will refactor the code, or maybe the rule will add compatibility support for this case), but I imagine it would be helpful for others as well.

#### Reproduction

https://github.com/Zip753/arrow-return-shorthand-migration-repro

KingDarBoja:
## PR Checklist

-   [x] Addresses an existing issue: fixes #345

## Overview

Provide a notice for the incompatible case commented on linked issue.

### DIFF
--- a/src/rules/converters/arrow-return-shorthand.ts
+++ b/src/rules/converters/arrow-return-shorthand.ts
@@ -1,5 +1,8 @@
 import { RuleConverter } from "../converter";
 
+export const ARROW_RETURN_NOTICE =
+    "ESLint will throw an error if the function body is multiline yet has one-line return on it.";
+
 export const convertArrowReturnShorthand: RuleConverter = (tslintRule) => {
     return {
         rules: [
@@ -9,6 +12,7 @@ export const convertArrowReturnShorthand: RuleConverter = (tslintRule) => {
                         ruleArguments: ["always"],
                     }),
                 ruleName: "arrow-body-style",
+                notices: [ARROW_RETURN_NOTICE],
             },
         ],
     };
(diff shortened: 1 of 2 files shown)


src/rules/converters/arrow-return-shorthand.ts
```diff
@@ -1,5 +1,8 @@
 import { RuleConverter } from "../converter";
 
+export const ARROW_RETURN_NOTICE =
+    "ESLint will throw an error if the function body is multiline yet has one-line return on it.";
```
JoshuaKGoldberg, line 4: Sorry to nitpick on grammar, just a small change...

```suggestion
    "ESLint will throw an error if the function body is multiline yet has a one-line return on it.";
```

PR #707, opened 2020-09-08
Defaulted --comment globs to --typescript settings

### LINKED ISSUE #426: Allow tsconfig files and --typescript to inform --comment
### 🚀 Feature Request

Splitting out followup work from #246: it'd be easier for users to run `--comment` if it could use tsconfig file inclusions to fill in the list of files to convert.

#### Existing Behavior

Right now you have to manually pass globs to `--comment`

#### Change Proposal

If `--comment` is provided as `true` (no globs are passed to it) and a `tsconfig.json` is available, it should use TypeScript logic to figure out the file inclusion list from there.

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #426

## Overview

If `--comments` is provided as `true` from having no globs passed to it, we'll now default to a configuration provided by `--typescript` if available. That necessitates passing in a found TypeScript configuration to `convertComments`.

Uses `minimatch` (already a transitive runtime dependency) to emulate TypeScript's `exclude`, `files`, and `include`.


KingDarBoja:
LGTM 🚀 But I think it should support specifying a custom `tsconfig` file path as some projects would use a different path or inherit from a base option which is not fully complete (like monorepos or Angular ones).

KingDarBoja (approved):

JoshuaKGoldberg:
YES! Excellent idea, will file a followup issue! 👏

PR #713, opened 2020-09-11
Allowed passing a tsconfig.json to --comments

### LINKED ISSUE #709: Allow passing a tsconfig.json path to --comments
### 🚀 Feature Request

As mentioned by @KingDarBoja in https://github.com/typescript-eslint/tslint-to-eslint-config/pull/707#issuecomment-688904736, some projects would use a different path or inherit from a base option which is not fully complete (like monorepos or Angular ones).

#### Existing Behavior

`--comments` can only be `true` to use the tsconfig set in `--typescript`, or a bunch of glob strings to include those files.

#### Change Proposal

If `--comments` is provided a single string _and_ that string ends with `.json`, we should consider that to be a path to a `tsconfig.json` and use it as the source of which files to replace comments in.

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #709

## Overview

As described by the issue, allows a string that ends with `.json` to be counted as a new TypeScript configuration to read from disk. In theory we could reuse the existing configuration if it's the same as `--typescript`, but that seems like both a rare case and not a significant performance penalty.

Extracts the logic to take `--comments` globs & existing TS config into a new injected function. This wasn't strictly necessary but they were getting pretty chunky.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ We **strongly** advise reading [docs/FAQs.md](./docs/FAQs.md) before planning yo
 
 Each of these flags is optional:
 
--   **[`comments`](#comments)**: File glob path(s) to convert TSLint rule flags to ESLint within.
+-   **[`comments`](#comments)**: TypeScript configuration or file glob path(s) to convert TSLint rule flags to ESLint within.
 -   **[`config`](#config)**: Path to print the generated ESLint configuration file to.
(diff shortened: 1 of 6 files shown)


KingDarBoja (commented):
Eveything looks great but the readme should be updated to include this change.

KingDarBoja (approved):
Perfection 👯‍♂️ 🍻

PR #724, opened 2020-09-22
feat: first approach to find existing eslint config

### LINKED ISSUE #722: Auto-detect original ESLint file in any form
### 🚀 Feature Request

Following #58, https://eslint.org/docs/user-guide/configuring states:

> **Configuration Files** - use a JavaScript, JSON or YAML file to specify configuration information for an entire directory and all of its subdirectories. This can be in the form of an `.eslintrc.*` file or an `eslintConfig` field in a `package.json` file, both of which ESLint will look for and read automatically, or you can specify a configuration file on the command line.

#### Existing Behavior

`tslint-to-eslint-config` will only look at `.eslintrc.js` by default if not explicitly told a path.

https://github.com/typescript-eslint/tslint-to-eslint-config/blob/2b48234f58b2fd75f3baf5017692c79d79a70942/src/cli/runCli.ts#L44

#### Change Proposal

Add a function dependency to `runCli` that determines the initial script path for ESLint, such as `.eslintrc.js`, `.eslintrc.yml`.

The tricky part here will be supporting `package.json` `eslintConfig`...

KingDarBoja:
## PR Checklist

-   [x] Addresses an existing issue: fixes #722

### DIFF
--- a/src/input/findExistingESLintConfiguration.ts
+++ b/src/input/findExistingESLintConfiguration.ts
@@ -1,0 +1,40 @@
+import { fsFileSystem } from "../adapters/fsFileSystem";
+import { GlobAsync } from "../adapters/globAsync";
+import { ResultStatus, ResultWithDataStatus } from "../types";
+import { ESLintConfiguration } from "./findESLintConfiguration";
+
+export type FindExistingESLintConfigurationDependencies = {
+    globAsync: GlobAsync;
+};
+
+export const findExistingESLintConfiguration = async (
+    dependencies: FindExistingESLintConfigurationDependencies,
+): Promise<ResultWithDataStatus<ESLintConfiguration | string[]>> => {
+    const globResult = await dependencies.globAsync(".eslintrc.{js,json,yaml}");
+    const packageContent = await fsFileSystem.readFile("package.json");
+    if (!(globResult instanceof Error)) {
+        return {
+            data: globResult,
+            status: ResultStatus.Succeeded,
+        };
+    }
+    if (globResult instanceof Error && packageContent instanceof Error) {
+        return {
+            errors: [globResult, packageContent],
+            status: ResultStatus.Failed,
+        };
+    }
+    if (globResult instanceof Error && typeof packageContent === "string") {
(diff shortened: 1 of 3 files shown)


src/input/findExistingESLintConfiguration.ts
```diff
@@ -0,0 +1,40 @@
+import { fsFileSystem } from "../adapters/fsFileSystem";
+import { GlobAsync } from "../adapters/globAsync";
+import { ResultStatus, ResultWithDataStatus } from "../types";
+import { ESLintConfiguration } from "./findESLintConfiguration";
+
+export type FindExistingESLintConfigurationDependencies = {
+    globAsync: GlobAsync;
+};
+
+export const findExistingESLintConfiguration = async (
```
JoshuaKGoldberg, line 10: Heh, yeah, the return type of this is what I was curious about... At first, I was thinking thinking the calling code will want to know whether this is an npm package style file or an ESLint one. For example, what happens if someone does something like `--config ./packages/somewhere/package.json`, and wants this to work on the `eslintConfig` in that nested file?

But, it's really only `package.json` files that have the odd `eslintConfig` member... Maybe the writing code would always want to override an existing `eslintConfig` in the output file if that output file's name ends with `"package.json"`? What do you think

src/input/findExistingESLintConfiguration.ts
```diff
@@ -0,0 +1,40 @@
...
+            status: ResultStatus.Succeeded,
+        };
+    }
+    if (globResult instanceof Error && packageContent instanceof Error) {
+        return {
+            errors: [globResult, packageContent],
+            status: ResultStatus.Failed,
+        };
+    }
+    if (globResult instanceof Error && typeof packageContent === "string") {
```
JoshuaKGoldberg, line 27: Nit, just marking this down because it confused me at first:

```suggestion
    if (typeof packageContent === "string") {
```

JoshuaKGoldberg:
Closing to keep the PR queue small. fyi @KingDarBoja 👍 thanks for the work on mapping out a first approach!

PR #728, opened 2020-10-02
Opt-in Atom editor conversion

### LINKED ISSUE #289: Support Atom editor setting conversion
### 🚀 Feature Request

#250 added an `--editor` feature to convert VS Code's TSLint extension settings to the VS Code ESLint equivalents. It'd be lovely to have a similar option for Atom: if a settings file for it exists, it should be converted.

#### Existing Behavior

`tslint-to-eslint-config` only looks for and supports the VS Code syntax for now.

#### Change Proposal

Whatever Atom uses ([this](https://flight-manual.atom.io/using-atom/sections/basic-customization/)?) should be supported as well. If _either_ config file exists, its respective conversion should be attempted.

### LINKED ISSUE #745: Switch editor config conversions to opt-in
### 🚀 Feature Request

Forking from #742: should editor config conversions not be applied if `--editor` isn't specified?

#### Existing Behavior

If a `.vscode/settings.json` exists, `tslint-to-eslint-config` will run on it.

#### Change Proposal

This can be surprising to users: they may not want to convert their editor config, and prior to #742 and #744 being resolved, this can be a destructive operation. Should we make conversion...

* ...opt-in until those two issues are resolved?
* ...opt-in even after those two issues are resolved?

Personally, I'm in favor of keeping the existing behavior because both those issues are pretty straightforward to resolve: there are already packages for [JSON5](https://json5.org) and [JSON diff patching](https://github.com/benjamine/jsondiffpatch). But we should discuss a bit instead of letting me steamroll 😄. What do you think, @KingDarBoja and @SgtPooki?

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #289; fixes #745 

## Overview

Builds off the excellent work by @MrCube42 (#133 -> #250) to make a general `convertEditorConfigs` function, which calls `convertEditorConfig` for each requested `--editor` configuration. Instead of assuming VS Code for all editors, a `editorConfigDescriptors` dependency is added that lists the known file path formats and matches to a known format if possible.

Trims down the logic around retrieving the settings keys and such to be a bit more manual in the editor configs. It was a nice thing to build off of but:
* Atom's editor settings such as [linter-eslint](https://atom.io/packages/linter-eslint) and [linter-tslint](https://atom.io/packages/linter-tslint) use nested properties, which are more difficult to support dynamically.
* The relevant editor extensions ended up not adding many new features, so the infrastructure was less necessary than I'd thought.

Adds two runtime dependencies to `package.json`:

* `cson-parser`: CSON equivalent to the global `JSON` object.
* `lodash`: This was already depended on at runtime so there's no real cost to using its nice `merge` function for merging configs... though this might be removed later in resolving #744. I'd like to make that change separately, to keep changes small.

Example output snippets from running `node ../tslint-to-eslint-config/bin/tslint-to-eslint-config --editor .vscode/settings.json`:

```
✨ 1 editor file augmented with its ESLint equivalent. ✨

❓ 1 vscode/settings.json editor setting is not known by tslint-to-eslint-config to have an ESLint equivalent. ❓
  Check ./tslint-to-eslint-config.log for details.
```

```
2 rules are not known by tslint-to-eslint-config to have ESLint equivalents:
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "typedef".
  * tslint-to-eslint-config does not know the ESLint equivalent for TSLint's "whitespace".
```


src/converters/editorConfigs/converters/convertVSCodeConfig.ts
```diff
@@ -0,0 +1,49 @@
...
+import { parseJson } from "../../../utils";
+import { EditorConfigConverter } from "../types";
+
+const knownMissingSettings = [
+    "tslint.alwaysShowRuleFailuresAsWarnings",
+    "tslint.exclude",
+    "tslint.ignoreDefinitionFiles",
+    "tslint.jsEnable",
+    "tslint.suppressWhileTypeErrorsPresent",
+];
```
KingDarBoja, line 13: Clever approach 🍰

KingDarBoja (approved):
Everything looks great to me, although I wish there was a easier way to spot changes when renaming files 😆

package.json
```diff
@@ -12,8 +12,10 @@
     "dependencies": {
         "chalk": "4.1.0",
         "commander": "6.1.0",
+        "cson-parser": "^4.0.5",
         "eslint-config-prettier": "6.12.0",
         "glob": "7.1.6",
+        "lodash": "^4.17.20",
```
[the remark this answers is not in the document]
JoshuaKGoldberg: Aha! I can never keep track of which Lodash is the right one to use now 😄. Will do.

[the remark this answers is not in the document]
JoshuaKGoldberg: Hmm, https://www.npmjs.com/package/lodash-es is a year out of date 😕 ... I'll skip this for now, but if it ends up getting regularly published it'd be great to include.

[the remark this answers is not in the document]
JoshuaKGoldberg: Oh fun, this is an actively discussed topic: https://github.com/lodash/lodash/issues/4879

package.json
```diff
@@ -12,8 +12,10 @@
     "dependencies": {
         "chalk": "4.1.0",
         "commander": "6.1.0",
+        "cson-parser": "^4.0.5",
```
skovalyov, line 15: I get an error `npm ERR! 404 Not Found - GET http://my-registry-url/coffeescript/-/coffeescript-1.12.8.tgz` during installation. It comes from `cson-parser`. There are some mentions of this problem, for example at jashkenas/coffeescript/issues/4805, but it's not clear how to fix it while installing Angular Eslint with `ng add @angular-eslint/schematics`

KingDarBoja: Hello, can you open a new issue in order to track the error with Angular ESLint + this package? Thank you :)

skovalyov: - https://github.com/angular-eslint/angular-eslint/issues/194
- https://github.com/typescript-eslint/tslint-to-eslint-config/issues/839

KingDarBoja: Thank you, we will look into these and provide a proper fix soon.

PR #731, opened 2020-10-03
Changed async addPrettierExtensions to synchronous checkPrettierExtension

### LINKED ISSUE #730: addPrettierExtensions doesn't need to be async
### Overview

Quick little tech debt point I just noticed while investigating #729: although other logic around rule extensions called by `summarizePackageRules` does need to be `async`, `addPrettierExtensions` itself makes no asynchronous calls. It can be changed to a non-`async` function.

The name also is a relic, I think, of an implementation draft that is no longer accurate. Maybe... `checkPrettierExtension` would be a better name for it? It doesn't add any extensions; it just checks whether the Prettier config _should_ be extended from.

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #730

## Overview

Does exactly what the issue suggested.


KingDarBoja (approved):
Tasty 🍖

PR #732, opened 2020-10-04
Added Node API

### LINKED ISSUE #729: Provide node API for config conversion logic
#### Existing Behavior

The library can be leveraged via a CLI and interacts directly with disk.

#### Change Proposal

Provide the option of leveraging (part of) the library in code via a node API which would only focus on the config conversion process.

After discussing this with @JoshuaKGoldberg it seems we essentially want to expose (potentially after some tweaks) the `convertLintConfig` function: https://github.com/typescript-eslint/tslint-to-eslint-config/blob/e425bab57c2b7ac84621abdfd86b3e6954ca9948/src/converters/lintConfigs/convertLintConfig.ts#L22

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #729 

## Overview

Creates a new `convertTSLintConfigStandalone` method that's exported as `convertTSLintConfig` by `src/index.ts`, along with all the `src/types` helpers used by its public API. Its usage is explained in a new `docs/API.md` file:

* It takes in the settings fields around config file locations and whether to use Prettier.
* It outputs both a raw object and formatted string for the output ESLint configuration.

This method necessitated making the first half of the existing `convertLintConfig` available on their own, so it's split out into a new `createESLintConfiguration` method.

I also moved all the dependency initialization to its own `src/api/dependencies.md` file so it can be re-used in that standalone method.


docs/API.md
```diff
@@ -0,0 +1,65 @@
+# API
+
+You can use `tslint-to-eslint-config` programmatically in your Node apps.
+It provides a **[`convertTSLintConfig`](#convertTSLintConfig)** function to find relevant configurations on disk and output the generated ESLint configuration.
+
+## `convertTSLintConfig`
```
JoshuaKGoldberg, line 6: @JamesHenry is this API the kind of thing you're looking for?

If you're expecting the Angular toolchain to also convert comments and editor configs, we could also, say, export those individually, along with a general catch-all `convert` function to run all the conversions. Otherwise I'd default to filing that as a followup issue.

JoshuaKGoldberg: Alternately, if some or all configs have already been loaded in, the relevant portions of `convertTSLintConfig` could be exported as well, to allow external tooling to piece it together...

JoshuaKGoldberg: Discussed in a 1:1: we'll need to export the relevant portions of the API. The full `convertTSLintConfig` reads to and writes from the file system, but the nx tooling this PR is intended for is all in-memory. Excellent.

JoshuaKGoldberg:
This should now be available on npm to try out locally via https://www.npmjs.com/package/tslint-to-eslint-config/v/2.0.0-beta0. 🚀

cc @JamesHenry

JamesHenry:
Thanks so much for publishing that @JoshuaKGoldberg!

After some digging I was able to get to the bottom of why my conversion was not behaving the same as when using the CLI (the rule severities and configs were not being reflected if you remember).

It is because `full` and `raw` are not equivalents, which is what I had been treating them as (in fact your code and docs in this PR currently also does).

Full is the result of running `tslint --print-config` and it is this JSON result which is what contains the necessary data about the rule severity and config which is then used when converting the config.

So I have managed to get it fully working after extracting some more parts of the library into my code.

I believe, however, I can avoid the need for that if you just expose to me the `findReportedConfiguration()` utility.

Here is effectively my usage:

```ts
const reportedConfiguration = await findReportedConfiguration<any>(
  childProcessExec,
  'npx tslint --print-config',
  pathToTslintJson
);

// ... error handling here...

const originalConfigurations = {
  tslint: {
    full: reportedConfiguration,
    raw: rawTslintJson,
  },
};
```

Let me know if you have any questions!

JamesHenry:
Thanks for the updates! Please could you publish another beta whenever you get chance? Not urgent!

JoshuaKGoldberg:
Super, published `2.0.0-beta1` @JamesHenry. 👍

JamesHenry:
This is looking really good! I haven't yet tackled the disable comment conversion logic, but for configs I think we are basically there. You can see my full usage of the API below for reference.

I have the following main request/callout as things stand:

**With this usage below I am still ending up with an empty `tslint-to-eslint-config.log` being created on disk.**

- I can of course manually clean this up, but the way that schematics work is that the user will be able to see that that file existed and that it was deleted as part of the run (which is very confusing cause it was not in there when they ran the command) [see screenshot]
  - image

- **Please can we hunt down that `createWriteStream` and prevent it from being executed when using the Node API?**

---

```ts
import type { Linter as ESLintLinter } from 'eslint';
import {
  createESLintConfiguration,
  findReportedConfiguration,
  joinConfigConversionResults,
} from 'tslint-to-eslint-config';

type TSLintRuleSeverity = 'warning' | 'error' | 'off';
type TSLintRuleOptions = {
  ruleArguments: any[];
  ruleName: string;
  ruleSeverity: TSLintRuleSeverity;
};

export async function convertToESLintConfig(
  pathToTslintJson: string,
  tslintJson: Record<string, unknown>,
): Promise<{
  convertedESLintConfig: ESLintLinter.Config;
  unconvertedTSLintRules: TSLintRuleOptions[];
}> {
  const reportedConfiguration = await findReportedConfiguration(
    'npx tslint --print-config',
    pathToTslintJson,
  );

  if (reportedConfiguration instanceof Error) {
    if (
      reportedConfiguration.message.includes('unknown option `--print-config')
    ) {
      throw new Error(
        '\nError: TSLint v5.18 required in order to run this schematic. Please update your version and try again.\n',
      );
    }
    throw new Error(`Unexpected error: ${reportedConfiguration.message}`);
  }

  const originalConfigurations = {
    tslint: {
      full: reportedConfiguration,
      raw: tslintJson,
    },
  };

  const summarizedConfiguration = await createESLintConfiguration(
    originalConfigurations,
  );

  const convertedESLintConfig = joinConfigConversionResults(
    summarizedConfiguration,
    originalConfigurations,
  ) as ESLintLinter.Config;

  return {
    convertedESLintConfig,
    unconvertedTSLintRules: summarizedConfiguration.missing,
  };
}
```

JoshuaKGoldberg:
Indeed, we can: the issue, I think, is that `processLogger.ts` creates a write stream immediately:

```
const debugFileName = "./tslint-to-eslint-config.log";

export const processLogger = {
    debugFileName,
    info: fs.createWriteStream(debugFileName),
```

...even though it's not used in this code. I'll switch it to only creating that write stream once, when first needed in calling `.info()`.

JoshuaKGoldberg:
@JamesHenry try the latest beta2?

JamesHenry:
beta2 is looking good!

JoshuaKGoldberg:
Informal approvals from @KingDarBoja (over Gitter) and @JamesHenry (by virtue of #173 existing and me presuming), so I'm going to merge this! 🚀 

If anybody has further requests on it, please do let me know somewhere and/or file an issue -- I'd like to tackle them before we stable release 2.0.0.

PR #816, opened 2020-11-08
Exported lint rule types for Node APIs

### LINKED ISSUE #814: Separately export (E|T)TSLintRuleOptions, (E|T)SLintRuleSeverity
### Overview

https://github.com/angular-eslint/angular-eslint/pull/173#discussion_r519441187: the Angular schematic/integration redeclares `TSLintRuleOptions` and `TSLintRuleSeverity` to strongly type the returns of its `convertToESLintConfig`.

#### Actual Behavior

These types are declared internally but never exported, so outside folks can't use them.

#### Expected Behavior

Export them via `src/index.ts` so Node API users can use them.

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #814

## Overview

Adds comments to and exports all the types in `src/lintConfigs/types.ts` referring to raw and reported lint rule configurations/options.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,6 +19,7 @@ module.exports = {
     plugins: ["@typescript-eslint"],
     rules: {
         "@typescript-eslint/require-array-sort-compare": "off",
+        "@typescript-eslint/comma-dangle": "off",
         "@typescript-eslint/consistent-type-definitions": ["error", "type"],
         "@typescript-eslint/consistent-type-imports": "off",
         "@typescript-eslint/explicit-function-return-type": "off",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,7 +19,6 @@ module.exports = {
     plugins: ["@typescript-eslint"],
     rules: {
         "@typescript-eslint/require-array-sort-compare": "off",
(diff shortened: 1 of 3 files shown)


JamesHenry:
Thanks! It's sort of tangential - but those objects represented by the `TSLintRuleOptions` interface are only available in TSLint via the CLI right (`--print-config`), there isn't an equivalent function or set of functions that can be called by importing them from 'tslint'?

Reason being, right now I am having to jump through some hoops in tests for the schematic because of the need to invoke that CLI within the implementation details, would be nice if there was a `printConfig()` util available!

JoshuaKGoldberg:
Hmm, good question. In theory it should be possible to export `printConfiguration` or something up the chain from it: https://github.com/palantir/tslint/blob/bfaf3e2e237f67e2eb6b28eab00e109a68602716/src/runner.ts#L179

...though there's nothing I can find in the current versions of TSLint that folks are using now 😞 .

JamesHenry:
Yeah I tried to trace it through at the time as well, and concluded it was non-trivial to expose it in a programmatic way. All good I've got unit tests working fine in that PR thanks to some jest mocking so it's not worth the hassle of taking it any further I don't think

PR #819, opened 2020-11-08
Split dependencies into their owning functions' files

### LINKED ISSUE #818: Investigate moving bound dependencies into their functions' files
### Overview

Idea from looking through #732: `processLogger` needed to be changed to lazily create the write stream... but isn't even used by the Node API. Although it'd be very hard to not import the file at all for the Node API if everything goes through `src/index.ts`, it'd still be nice to not set up its dependencies unnecessarily.

#### Actual Behavior

All dependencies are set up in a single `src/api/dependencies.ts` file right now.

#### Expected Behavior

Can each function's file import its own dependencies set up and export the function's dependencies object?

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #818

## Overview

Instead of declaring all dependency objects in a single `src/api/dependencies.ts`, moves their declarations to be next to the functions that use them.


docs/Dependencies.md
```diff
@@ -1,3 +1,5 @@
+# TODO: REWRITE IF CHANGES DEEMED ACCEPTABLE
```
JoshuaKGoldberg, line 1: @KingDarBoja do you have any thoughts on the changes here?

KingDarBoja: I have nothing against it, as long as there aren't API changes but only moving those dependency definitions into each owning directories. Based on that, pretty much point out that now eachdependency lives on the same file as the function fiels. ⭐

JoshuaKGoldberg:
Closing to keep my PR queue small 😄

PR #821, opened 2020-11-08
Added convertFileComments API

### LINKED ISSUE #820: Also export comment converter Node APIs
### 🚀 Feature Request

Following #732 to add a Node API, the next step for https://github.com/angular-eslint/angular-eslint/pull/173 will be to also convert inline `tslint:disable` comments for users. 

#### Existing Behavior

The `convertFileComments` method is not exported as a standalone function via the Node API.

#### Change Proposal

Let's export it!

More specifically: let's export an equivalent to it that takes in a file's content and path, and returns the new equivalent.

JoshuaKGoldberg:
## PR Checklist

-   [x] Addresses an existing issue: fixes #820

## Overview

Adds a synchronous single-file-use function to convert a file's contents as a string into its ESLint equivalent.

```ts
import { convertFileComments } from "tslint-to-eslint-config";

// "// eslint-disable-next-line"
const newContents = convertFileComments({
    fileContent: "// tslint:disable-next-line",
    filePath: "a.ts",
});
```


docs/API.md
```diff
@@ -136,3 +136,59 @@ const raw = joinConfigConversionResults(summarizedConfiguration, originalConfigu
...
+
+## Standalone Comment Conversion API
+
+> ⚠ This area of code is still considered experimental.
+> Use at your own risk.
+> Please file an issue on GitHub if you'd like to see changes.
+
+The individual per-file conversion logic within `convertComments` is available as a standalone function:
+
+### `convertFileComments`
```
JoshuaKGoldberg, line 148: @JamesHenry this is published under v2.0.0-beta4 now, if you want to try it out. Is this the kind of API you're looking for?

I'd recommend using both `ruleCommentsCache` (for a small perf boost) and `ruleEquivalents` (for rule accuracy per arguments).

JamesHenry:
Thanks so much for the super fast turnaround! On the face of it this is exactly what I need. It's getting late here but I'll definitely incorporate this into the PR this week, thanks again!

JamesHenry:
Hmm just installing beta4 and running a build/link with lerna is blowing up

image

I see the install affected TS versions in my yarn.lock file

image

I wonder if that could be related...

JamesHenry:
Confirmed also happens with beta3, but not when restoring to beta2 which is where I was before

JoshuaKGoldberg:
Oh you know what, I probably neglected to remove `.d.ts` files left over from old file names, manually published locally, and now they're checked into npm. Glorious. This is why we should have automated package publishing 😄 .

Published a beta5. If that still has the `.d.ts` cruft you can delete them locally (I'd also suggest changing your tsc settings to not pick up on those orphan node_modules/ files).

JamesHenry (approved):
This seems to be working great for my use-case! Thanks again

docs/API.md
JamesHenry: Ooh I missed this tip and just used your snippet in the description, I'll look at adding them in

PR #924, opened 2021-01-10
refactor: remove use-pipe-decorator converter

rafaelss95:
## Overview
As mentioned in https://github.com/mgechev/codelyzer/issues/592, https://github.com/angular-eslint/angular-eslint/pull/241 and https://github.com/angular-eslint/angular-eslint/pull/245, this rule seems to be unnecessary since Angular 8 (pre-Ivy) and in order to drop it in angular-eslint repo the converter should be removed first.

cc @JamesHenry @JoshuaKGoldberg.


JoshuaKGoldberg (approved):
Thanks @rafaelss95!

PR #1062, opened 2021-05-16
chore: update typescript-eslint to v4.23.0

KingDarBoja:
## Overview

Should fix #1035 CI issues due to ESLint rules complaints.


JoshuaKGoldberg (approved):
😍 !

PR #1085, opened 2021-05-22
Add SonarTS converters.

Res42:
## PR Checklist

-   [X] Addresses an existing issue: fixes #1083
-   [X] That issue was marked as [`status: accepting prs`](https://github.com/typescript-eslint/tslint-to-eslint-config/labels/status%3A%20accepting%20prs)

## Overview

I added the rules from the table.

I have some questions:

1. There are existing converters that does the same as a new one for SonarTS. Should I use the existing converter or are these distinct business logic?

| Existing converter | New converter | ESLint rule |
| --- | --- | --- |
| convertAdjacentOverloadSignatures | convertConsecutiveOverloads | @typescript-eslint/adjacent-overload-signatures |
| convertNoUnnecessarySemicolons | convertNoExtraSemicolon | no-extra-semi |
| convertNoForInArray | convertNoInMisuse | @typescript-eslint/no-for-in-array |
| convertAwaitPromise | convertNoInvalidAwait | @typescript-eslint/await-thenable |
| convertNoMultilineString | convertNoMultilineStringLiterals | no-multi-str |
| convertNoUnnecessaryTypeAssertion | convertNoUselessCast | @typescript-eslint/no-unnecessary-type-assertion |
| convertNoConstruct | convertUsePrimitiveType | no-new-wrappers |

2. When do you need mergers? These are the new converters that already have an existing one, but with different rule arguments:

| Existing converter | New converter | ESLint rule |
| --- | --- | --- |
| convertNoEmpty | convertNoEmptyNestedBlocks | no-empty |
| convertNoUseBeforeDeclare | convertNoVariableUsageBeforeDeclaration | @typescript-eslint/no-use-before-define |
| convertBanTypes | convertUsePrimitiveType | @typescript-eslint/ban-types |

Also in the new converter `convertUsePrimitiveType` I use the ESLint rule `@typescript-eslint/ban-types` with the types `String`, `Boolean`, and `Number`. Should I do this? I don't know if it is better if I just use the default options (by not giving any arguments) or if the merger can handle if this converter uses these explicit types and another converter use the default (no args) options.

[...]

Issue #813, opened 2020-11-07
simpleRuleConverter and argument helpers for simple cases 


criskrzysiu:
### 🚀 Feature Request

1. I think it is tedious to create all the files, tests, etc... for simple rule converters that are basically changing the name, adding plugins and that's it. For such plugins, it would be great to add a simple helper function or even an additional place that we just pass those simple values.

2.  For simple cases of argument mapping we could add a mapping function that would help with that. 

#### Existing Behavior

Even a simple rule takes at least 2 new files and 1 change in the existing one - often copy-pasting code. 

#### Change Proposal

PR in progress

JoshuaKGoldberg:
👋  @criskrzysiu are you still interested in tackling this? No worries if not, I can if you don't have time!

JoshuaKGoldberg:
I'm going to close this issue for aging out. Now that we have a utility script to auto-generate those templating files (#1089) I'm not convinced this is worth the effort.

Please post back here if you disagree!

PR #1089, opened 2021-05-23
Added quick utility script for making a new rule

JoshuaKGoldberg:
@Res42's efforts in #1085 reminded me I've been meaning to write this for a little while... should be helpful in setting up new rules, in lieu of #813. Thoughts?


script/new-converter.js
```diff
@@ -0,0 +1,63 @@
...
+                {
+                    ruleName: "${args.eslint}",
+                },
+            ],
+        });
+    });
+});
+`.trimLeft(),
+    );
+})();
```
JoshuaKGoldberg, line 63: This script doesn't (yet?) add to `ruleConverters.ts`. I ran out of energy :smile: but it would be a good followup... or I might tackle it soon.

Res42:
I generated my rules too, so here are some thoughts.

script/new-converter.js
```diff
@@ -0,0 +1,63 @@
...
+    for (const arg of ["eslint", "tslint"]) {
+        if (!args[arg]) {
+            throw new Error(`Missing --${arg} option.`);
+        }
+    }
+
+    const tslintPascalCase = upperFirst(camelCase(args.tslint));
+    const plugins = args.eslint.includes("/")
+        ? `
+        plugins: ["${args.eslint.split("/")[0]}"],`
```
Res42, line 21: Currently there is a plugin that has a prefix and it is not included in the plugin list in converters: `@typescript-eslint`. It is assumed that plugin is always installed I think?

JoshuaKGoldberg: Ha, true. I was just thinking of the react-a11y ones!

script/new-converter.js
JoshuaKGoldberg: I will after merging in from #1102

KingDarBoja (approved):
LGTM, and I trust it works even if there are no tests 😆

PR #1092, opened 2021-05-23
Do not use merger if arguments are the same.

### LINKED ISSUE #1086: Don't bother with a rule merger if rule arguments are the same
### Overview

As noted in #1085: some community rulesets have overlapping rules to others with the same output ESLint rules: e.g. both `convertAdjacentOverloadSignatures` and `convertConsecutiveOverloads` outputting `@typescript-eslint/adjacent-overload-signatures`.

#### Actual Behavior

A rule merger will be requested the second+ time an ESLint rule is seen in a converter's output: even if it outputs the same rule arguments as the ESLint rule's previously output rule arguments.

#### Expected Behavior

A rule merger should only be called for if an ESLint's rules are seen a second+ time _with new rule arguments_.

Res42:
## PR Checklist

-   [X] Addresses an existing issue: fixes #1086
-   [X] That issue was marked as [`status: accepting prs`](https://github.com/typescript-eslint/tslint-to-eslint-config/labels/status%3A%20accepting%20prs)

## Overview

Do not use merger if arguments are the same.

Also this adds a new feature: creating a merger is not needed if there are no configuration options for the rule.

Removed unneeded mergers (for rules with no configuration options).

### DIFF
--- a/docs/Architecture/Linters.md
+++ b/docs/Architecture/Linters.md
@@ -52,6 +52,8 @@ These are located in `src/rules/mergers/`, and keyed under their names by the ma
 
 For example, `@typescript-eslint/ban-types` spreads both arguments' `types` members into one large `types` object.
 
+**A merger does not need to be created if the rule does not accept any configuration.**
+
 ## Package Summaries
 
 ESLint configurations are summarized based on extended ESLint and TSLint presets.
--- a/docs/Architecture/Linters.md
+++ b/docs/Architecture/Linters.md
@@ -52,7 +52,7 @@ These are located in `src/rules/mergers/`, and keyed under their names by the ma
 
 For example, `@typescript-eslint/ban-types` spreads both arguments' `types` members into one large `types` object.
 
-> A merger does not need to be created if the rule does not accept any configuration.
+> A merger does not need to be created if the rule does not accept any configuration or all converters output exactly the same configuration.
 
 ## Package Summaries
 
--- a/docs/Architecture/Linters.md
+++ b/docs/Architecture/Linters.md
@@ -52,7 +52,7 @@ These are located in `src/rules/mergers/`, and keyed under their names by the ma
 
 For example, `@typescript-eslint/ban-types` spreads both arguments' `types` members into one large `types` object.
 
-**A merger does not need to be created if the rule does not accept any configuration.**
+> A merger does not need to be created if the rule does not accept any configuration.
 
 ## Package Summaries
 
--- a/docs/Architecture/Linters.md
+++ b/docs/Architecture/Linters.md
@@ -23,6 +23,8 @@ Those are run by `src/converters/lintConfigs/rules/convertRules.ts`, which takes
     * The output rule name is added to the TSLint rule's equivalency set.
     * The TSLint rule's config severity is mapped to its ESLint equivalent.
     * If this is the first time the output ESLint rule is seen, it's directly marked as converted.
+    * Notices are merged and deduplicated.
+    * If the existing output has the same arguments as the new output, merge lookups are skipped.
     * If not, a rule merger is run to combine it with its existing output settings.
 
 ### Rule Converters
--- a/src/converters/lintConfigs/rules/convertRules.test.ts
+++ b/src/converters/lintConfigs/rules/convertRules.test.ts
@@ -107,41 +107,118 @@ describe("convertRules", () => {
         expect(ruleEquivalents).toEqual(new Map([["tslint-rule-a", ["eslint-rule-a"]]]));
     });
 
-    it("reports a failure when two outputs exist for a converted rule without a merger", () => {
-        // Arrange
-        const conversionResult = {
-            rules: [
-                {
-                    ruleName: "eslint-rule-a",
-                },
-                {
-                    ruleName: "eslint-rule-a",
-                },
-            ],
-        };
-        const { tslintRule, converters, mergers } = setupConversionEnvironment({
-            conversionResult,
+    describe("runs without an argument merger when both rules have the same arguments", () => {
+        [
+            { ruleArguments: undefined, testCase: "no arguments" },
+            { ruleArguments: [5], testCase: "number argument" },
+            { ruleArguments: ["allow-something"], testCase: "string argument" },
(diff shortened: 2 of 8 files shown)


docs/Architecture/Linters.md
```diff
@@ -52,6 +52,8 @@ These are located in `src/rules/mergers/`, and keyed under their names by the ma
 
 For example, `@typescript-eslint/ban-types` spreads both arguments' `types` members into one large `types` object.
 
+**A merger does not need to be created if the rule does not accept any configuration.**
```
JoshuaKGoldberg, line 55: A couple nits:
* The docs generally use `> ` quotes for slightly emphasized stuff, to be less emphatic.
* We'll probably end up in this situation:

```suggestion
> A merger does not need to be created if the rule does not accept any configuration, or all converters output the same settings.
```

src/converters/lintConfigs/rules/convertRules.test.ts
```diff
@@ -107,41 +107,118 @@ describe("convertRules", () => {
...
-            conversionResult,
+    describe("runs without an argument merger when both rules have the same arguments", () => {
+        [
+            { ruleArguments: undefined, testCase: "no arguments" },
+            { ruleArguments: [5], testCase: "number argument" },
+            { ruleArguments: ["allow-something"], testCase: "string argument" },
+            { ruleArguments: [{ max: 50 }], testCase: "object argument" },
+            { ruleArguments: [5, "allow-something", { max: 50 }], testCase: "multiple arguments" },
+        ].forEach(({ ruleArguments, testCase }) => {
+            it(testCase, () => {
```
JoshuaKGoldberg, line 118: I think this is what you're looking for ✨ : https://jestjs.io/docs/api#testeachtablename-fn-timeout

```ts
        [[5], "no arguments"],
        // ...
    ],
    test('%s', (ruleArguments)) => {
```

Though, I think we can safely reduce this down to just the `undefined` case and the last, complex one. Since it uses `isEqual` from Lodash we don't particularly need to test the equality checking logic. Thoughts?

src/converters/lintConfigs/rules/convertRules.ts
```diff
@@ -109,3 +116,10 @@ export const convertRules = (
 
     return { converted, failed, missing, plugins, ruleEquivalents };
 };
+
+function mergeNotices(existingConversion: ESLintRuleOptions, newConversion: ESLintRuleOptions) {
```
JoshuaKGoldberg, line 120: This is already implemented in `uniqueFromSources`. ✨

src/converters/lintConfigs/rules/convertRules.test.ts
```diff
@@ -201,7 +278,8 @@ describe("convertRules", () => {
         );
 
         // Assert
-        expect(converted).toEqual(
+        expect([
+            // don't care if arguments merger ran or not
```
JoshuaKGoldberg, line 282: Nit: I think we can do away with these two extra comment lines; they're implied by the tests' behavior.

```suggestion
```

src/converters/lintConfigs/rules/convertRules.ts
```diff
@@ -78,21 +80,26 @@ export const convertRules = (
                     continue;
                 }
 
-                // 4d. If not, a rule merger is run to combine it with its existing output settings.
+                // 4d. First it merges the notices, since that is always needed.
```
JoshuaKGoldberg, line 83: These intentionally reflect what's in `docs/Architecture/Linters.md`. Could you please update that file too?

Also, nit, phrases like "is always needed" are generally implied by the context.

```suggestion
                // 4d. First it deduplicates and merges notices
```

JoshuaKGoldberg (changes requested):
Core logic looks great -- and greatly appreciated on the thorough tests and merger cleanups! Requesting changes around the docs linking and style.

docs/Architecture/Linters.md
Res42: Sure, I'll use `>`, but I think your text modification is wrong.

The merger will not run if two output configuration are the same, but that is just an implementation detail.
A merger needs to exists if there is any configuration at all.

JoshuaKGoldberg: If the two converters output the _same_ configuration, we wouldn't want to use a merger. There'd be no need.

src/converters/lintConfigs/rules/convertRules.test.ts
Res42: I mainly use `jasmine` so I didn't know of this feature. I'll refactor these parametrized tests.

docs/Architecture/Linters.md
Res42: I understand what you mean, but:

My thought process was that when you add a new converter, then the workflow is:
1. If rule has no configuration: no merger needed. I wrote this in the docs.
2. Else
    1. If other converter(s) output the same rule, then a merger is needed, since you can't assume what configuration arguments will it output and if this arguments output will be the same as the new converter.
    2. Else no merger is needed.

An example for this is `2.i` case is a (fictional) converter like this:

```ts
import { RuleConverter } from "../../ruleConverter";

export const convertNoBigFunction: RuleConverter = (tslintRule) => {
    return {
        rules: [
            {
                ...(tslintRule.ruleArguments.length !== 0 && {
                    ruleArguments: tslintRule.ruleArguments,
                }),
                ruleName: "max-lines-per-function",
            },
        ],
    };
};
```

So I think most if the time *when writing a new converter* you don't know if the output rule arguments will be the same when actually running the program.

So I think you should not encourage not writing a merger when the rule has configuration and there is an existing converter.

Res42: Maybe it would be better to say that:

> A merger does not need to be created if the rule does not accept any configuration or neither converters outputs arguments.

JoshuaKGoldberg: That works for me.

Res42: I read your suggestion again and I think I misinterpreted it originally. I'll use your suggestion.

Res42:
Done all change requests. Should I ping now?

JoshuaKGoldberg (approved):
Fantastic, thanks so much @Res42!

Issue #149, opened 2019-09-10
Conflict between tslint:align and @typescript-eslint/indent


rhysowen9:
I'm running into a conflict between @typescript-eslint/indent and the align rule in @typescript-eslint/tslint/config, specifically with regard to the alignment of parameters on my class constructors.

### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 0.2.5
-   ESLint version: 6.3.0
-   Node version: 10.15.3

#### Actual Behavior

Expected indentation of x spaces but found x  @typescript-eslint/indent
or
parameters are not aligned (tslint:align)  @typescript-eslint/tslint/config

#### Expected Behavior

I don't really mind which convention is accepted as long as its consistent.

#### Reproduction

```ts
constructor(private router: Router,
            private authenticationService: AuthenticationService,
            private tokenService: TokenService) {}
```

```ts
constructor(private router: Router,
  private authenticationService: AuthenticationService,
  private tokenService: TokenService) { }
```

```JSON
{
    "rules": {
        "@typescript-eslint/indent": [
            "error",
            2
        ],
        "@typescript-eslint/tslint/config": [
            "error",
            {
                "rulesDirectory": [
                    "**MyPath**"
                ],
                "rules": {
                    "align": [
                        true,
                        "parameters",
                        "statements"
                    ]
                }
            }
        ]
    }
}
```

JoshuaKGoldberg:
Thanks for the report @rhysowen9! It sounds like the `align` rule needs a converter that outputs the `indent` rule. Accepting PRs to add that in.

[`docs/Development.md`](https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/docs/Development.md) has a root overview for how to get started. Within that, [`docs/Architecture.md#rule-converters`](https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/docs/Architecture.md#rule-converters) talks about how these converters operate.

KingDarBoja:
@JoshuaKGoldberg I do think it is already fixed because I see the rule on the `converters` map.

JoshuaKGoldberg:
@KingDarBoja the issue here is that TSLint's `align` rule isn't yet converted - not the `indent` one, sadly;.

KingDarBoja:
> @KingDarBoja the issue here is that TSLint's `align` rule isn't yet converted - not the `indent` one, sadly.

I will look into it these days 🍰

[...]

PR #1093, opened 2021-05-23
Fix @typescript-eslint disable rules.

### LINKED ISSUE #1091: Missing ESLint rule disable in @typescript-eslint rule converter(s)
### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 2.4.2
-   ESLint version: -
-   Node version: -

#### Actual Behavior

This issue was created according to this comment: <https://github.com/typescript-eslint/tslint-to-eslint-config/pull/1085#issuecomment-846487277>.

There are [@typescript-eslint/...](https://github.com/typescript-eslint/typescript-eslint) rules that require to turn off their `ESLint` counterpart like this:

```jsonc
{
  // note you must disable the base rule as it can report incorrect errors
  "no-use-before-define": "off",
  "@typescript-eslint/no-use-before-define": ["error"]
}
```
Source: <https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-use-before-define.md>

In the rule [convertNoUseBeforeDeclare](https://github.com/typescript-eslint/tslint-to-eslint-config/blob/main/src/converters/lintConfigs/rules/ruleConverters/no-use-before-declare.ts) this is not disabled.

Also I think it would be good to check all `@typescript-eslint` rule converters for this type of bug.

#### Expected Behavior

`@typescript-eslint` rule converters that require to turn off their `ESLint` rule counterpart actually turn them off.

Res42:
## PR Checklist

-   [X] Addresses an existing issue: fixes #1091
-   [X] That issue was marked as [`status: accepting prs`](https://github.com/typescript-eslint/tslint-to-eslint-config/labels/status%3A%20accepting%20prs)

## Overview

Fix converters for `@typescript-eslint` rules where the rule needs the base `ESLint` rule to be disabled.
Also added a sanity check test which *naively* searches for these rule pairs and fails if the base rule is not disabled.


src/converters/lintConfigs/rules/ruleConverters/align.ts
```diff
@@ -29,6 +29,10 @@ export const convertAlign: RuleConverter = (tslintRule) => {
 
     return {
         rules: [
+            {
+                ruleName: "indent",
+                ruleSeverity: "off",
+            },
```
KingDarBoja, line 35: Based on #149, I think it is safe to  update the [Typescript ESLint Roadmap](https://github.com/typescript-eslint/typescript-eslint/blob/61e035356ced5186040eb4aeb78bd74c1c1ce4b8/packages/eslint-plugin/ROADMAP.md) with this conversion. @JoshuaKGoldberg

KingDarBoja (approved):
Outstanding job! 🇭🇺

src/converters/lintConfigs/rules/meta.test.ts
```diff
@@ -0,0 +1,84 @@
+import { promises as fs } from "fs";
+import { resolve, join } from "path";
+
+describe("@typescript-eslint rules that require another rule be disabled", () => {
```
JoshuaKGoldberg, line 4: Err, I don't think this test needs to be written? A lot of the rules (e.g. `default-param-last`) weren't TSLint rulees and thus aren't relevant to this project. The rest are already tested in their own files.

JoshuaKGoldberg (changes requested):
The rule converter fixes are EXCELLENT, thank you! But: I don't think the added test needs to be there?

src/converters/lintConfigs/rules/meta.test.ts
Res42: I written these so I could find the buggy rules and thought if I already written it, it can stay.
If/when a new conversion is added which forgets to disable a rule from here a test will break.

But if you want I will delete it.

JoshuaKGoldberg (approved):
Thanks! 💯

PR #1201, opened 2021-08-30
feat: add missing angular-whitespace converter

rafaelss95:
## PR Checklist

-   [ ] Addresses an existing issue:

## Overview
**1st. commit**: simplify converters to simply return an empty object, so it's consistent with existing ones (sorry, I didn't realize this before);
**2nd. commit**: move `import-destructuring-spacing` and `prefer-inline-decorator` from root folder to `codelyzer/`;
**3rd. commit**: create a converter for `angular-whitespace`.

It's still missing the rule `no-unused-css`... I don't know if it should be added as it could be potentially be added someday (https://github.com/angular-eslint/angular-eslint/issues/451)... in any case, tell me what to do.


rafaelss95:
Probably will drop the 2nd. commit as per #1203 😆

JoshuaKGoldberg:
Apologies for the delay on these: expect a review by the end of the weekend!

JoshuaKGoldberg (approved):
Super, thanks! 🚀

JoshuaKGoldberg:
> `no-unused-css`

Yeah, that's a tricky one. I think it's fine to leave as-is. The current message of _"we don't have an equivalent for this rule"_ is accurate.

Issue #1028, opened 2021-04-22
[variable-name] with ban-keywords converted to deprecated ban-keywords instead of new id-denylist


yrtimiD:
### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 2.3.0
-   ESLint version: 7.24.0

#### Actual Behavior
```json
"variable-name": [
  true,
  "ban-keywords"
]
```
is converted to 
```json
"id-blacklist": [
  "error",
  "any",
  "Number",
  "number",
  "String",
  "string",
  "Boolean",
  "boolean",
  "Undefined",
  "undefined"
]
```
#### Expected Behavior
id-blacklist was deprecated in favor of id-denylist since ESLint 7.4.0

```json
"id-denylist": [
  "error",
  "any",
  "Number",
  "number",
  "String",
  "string",
  "Boolean",
  "boolean",
  "Undefined",
  "undefined"
]
```
should be simple rename, code is same

JoshuaKGoldberg:
Love it, thanks for the report @yrtimiD! Agreed this would be a great first issue to tackle for someone.

hyperupcall:
It looks like #1221 fixes this

JoshuaKGoldberg:
Ah good point, thanks for the spot!

PR #1221, opened 2021-09-18
fix: [variable-name] Use `id-denylist` ESLint rule instread of `id-blacklist` that deprecated

wafuwafu13:
## PR Checklist

-   [x] Addresses an existing issue: fixes https://github.com/typescript-eslint/tslint-to-eslint-config/issues/1028

## Overview

`id-blacklist` was deprecated in favor of `id-denylist` since ESLint 7.4.0.

I was able to confirm that the logic of the options is the same, so I just renamed it.

**id-blacklist**
https://github.com/eslint/eslint/blob/47be8003d700bc0606495ae42610eaba94e639c5/lib/rules/id-blacklist.js#L140-L153

**id-denylist**
https://github.com/eslint/eslint/blob/47be8003d700bc0606495ae42610eaba94e639c5/lib/rules/id-denylist.js#L121-L134

### DIFF
--- a/src/converters/lintConfigs/rules/ruleConverters/tests/variable-name.test.ts
+++ b/src/converters/lintConfigs/rules/ruleConverters/tests/variable-name.test.ts
@@ -29,7 +29,7 @@ describe(convertVariableName, () => {
                     ruleName: "no-underscore-dangle",
                 },
                 {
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg (approved):
💖 love it, thanks!

PR #1565, opened 2022-08-26
fix(rule/variable-name): Use `ruleArguments` property

### LINKED ISSUE #1225: variable-name generates naming-convention
### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 2.10.0
-   ESLint version: N/A
-   Node version: v14.17.6

#### Actual Behavior

Running tslint-to-eslint-config on a tslint.json file with a `variable-name` rule, but no "check-format" option:

```
{
    "rules": {
        "variable-name": [true, "ban-keywords"]
    }
}
```

Outputs an `.eslintrc.js` file with a default `naming-convention` rule (without a selector).

```
{
    ...
    "rules": {
        "@typescript-eslint/naming-convention": "error",
        "id-denylist": [
            "error",
            "any",
            "Number",
            "number",
            "String",
            "string",
            "Boolean",
            "boolean",
            "Undefined",
            "undefined"
        ],
        "id-match": "error",
        "no-underscore-dangle": "error"
    }
}
```

#### Expected Behavior

I would expect this rule to only output the `id-denylist` rule, or at worst output a `naming-convention` rule with the `variable` selector as seems to be specified in the [unit test](https://github.com/typescript-eslint/tslint-to-eslint-config/blob/bf5c9f9f03a67285d9eed40e2fccd86c4018f35d/src/converters/lintConfigs/rules/ruleConverters/tests/variable-name.test.ts).

#### Reproduction

[Minimal reproducible example repo](https://github.com/benjaminpjones/variable-name-test)
[Real life repo](https://github.com/online-go/online-go.com) - Running eslint with a `tslint-to-eslint-config`-generated `.eslintrc.js` generates a ton of new `naming-convention` errors, which I think can be avoided.

hyperupcall:
## PR Checklist

-   [x] Addresses an existing issue: fixes #1225

Hello again! :smile: 

## Overview

As mentioned in the linked issue, one problem was that the arguments for `"@typescript-eslint/naming-convention"` weren't being passed to the ESLint rule properly, so it just said `"error"`. From my understanding, this caused problems since this would also include lint errors for function names, etc. too

It seems the fix was just a typo in the plugin. I also added a test specifically for `ban-keywords`, so it is a little more clear what is going on. This code generates the following ESLint file (for the same `tslint.json` file in the linked issue):

<details>
<summary>.eslintrc</summary>

```js
Hi! This file was autogenerated by tslint-to-eslint-config.
https://github.com/typescript-eslint/tslint-to-eslint-config

It represents the closest reasonable ESLint configuration to this
project's original TSLint configuration.

We recommend eventually switching this configuration to extend from
the recommended rulesets in typescript-eslint.
https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/docs/FAQs.md

Happy linting! M-pM-^_M-^RM-^V
*/
module.exports = {
    "env": {
        "browser": true,
        "es6": true
    },
    "extends": [
        "prettier"
    ],
    "parser": "@typescript-eslint/parser",
    "parserOptions": {
        "project": "tsconfig.json",
        "sourceType": "module"
    },
    "plugins": [
        "@typescript-eslint"
    ],
    "root": true,
    "rules": {
        "@typescript-eslint/naming-convention": [
            "error",
            {
                "selector": "variable",
                "format": [
                    "camelCase",
                    "UPPER_CASE"
                ],
                "leadingUnderscore": "forbid",
                "trailingUnderscore": "forbid"
            }
        ],
        "id-denylist": [
            "error",
            "any",
            "Number",
            "number",
            "String",
            "string",
            "Boolean",
            "boolean",
            "Undefined",
            "undefined"
        ],
        "id-match": "error",
        "no-underscore-dangle": "error"
    }
};
```

</details>

Also it was stated in the linked issue that the best thing would possibly be "only outputing the `id-denylist`" rule - this doesn't address that part specifically, but does fix the bug within the rule.

To make sure this issue doesn't exist for other plugins, maybe it might be a good idea to validate whatever a plugin returns against some schema, at least in testing? That way, a plugin won't accidentally set `rules` instead of `ruleArguments` without some alarms going off.

### DIFF
--- a/src/converters/lintConfigs/rules/ruleConverters/tests/variable-name.test.ts
+++ b/src/converters/lintConfigs/rules/ruleConverters/tests/variable-name.test.ts
@@ -17,7 +17,7 @@ describe("convertVariableName", () => {
             rules: [
                 {
                     ruleName: "@typescript-eslint/naming-convention",
-                    rules: [
+                    ruleArguments: [
                         {
                             selector: "variable",
                             format: ["camelCase", "UPPER_CASE"],
@@ -49,7 +49,7 @@ describe("convertVariableName", () => {
             rules: [
                 {
                     ruleName: "@typescript-eslint/naming-convention",
-                    rules: [
+                    ruleArguments: [
                         {
                             selector: "variable",
                             format: ["camelCase", "UPPER_CASE"],
@@ -81,7 +81,7 @@ describe("convertVariableName", () => {
             rules: [
                 {
                     ruleName: "@typescript-eslint/naming-convention",
-                    rules: [
+                    ruleArguments: [
                         {
                             selector: "variable",
                             format: ["camelCase", "UPPER_CASE"],
@@ -113,7 +113,7 @@ describe("convertVariableName", () => {
             rules: [
                 {
                     ruleName: "@typescript-eslint/naming-convention",
-                    rules: [
+                    ruleArguments: [
                         {
                             selector: "variable",
                             format: ["camelCase", "UPPER_CASE", "PascalCase"],
@@ -145,7 +145,7 @@ describe("convertVariableName", () => {
             rules: [
                 {
                     ruleName: "@typescript-eslint/naming-convention",
-                    rules: [
+                    ruleArguments: [
                         {
                             selector: "variable",
                             format: ["camelCase", "UPPER_CASE"],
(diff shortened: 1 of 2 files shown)


JoshuaKGoldberg:
> maybe it might be a good idea to validate whatever a plugin returns against some schema, at least in testing? That way, a plugin won't accidentally set rules instead of ruleArguments without some alarms going off.

Agreed, that'd be great. If you have an idea for how you'd like to do it, I'd love an issue

JoshuaKGoldberg (approved):
Seems reasonable, thanks!

PR #1639, opened 2022-12-02
Improve JSON parse test assertion to work on recent Node versions

hyperupcall:
## Overview

I'm on Node `v19.0.0`, and it seems the JSON parser has been improved to be more descriptive. However, this causes a test to fail, since the `.message` property of the `SyntaxError` is slightly different. 

image

This PR fixes that by matching `.message` with a Regex directly. I tried to use `[...].toContainEqual(result)`, but that didn't seem to work, so this fix seemed to be the next best alternative.


JoshuaKGoldberg (approved):
Yeah this makes sense - thanks!

PR #1717, opened 2023-08-02
fix: Do not show Errors from reading directories

### LINKED ISSUE #1720: --comments throws EISDir when a directory is specified in TSConfig include
### 🐛 Bug Report

-   `tslint-to-eslint-config` version: 2.14.0
-   ESLint version: n/a
-   Node version: v20.2.0 (n/a)

#### Actual Behavior

```plaintext
$ npx tslint-to-eslint-config --comments

✨ 1 rule replaced with its ESLint equivalent. ✨

❗ 1 ESLint rule behaves differently from its TSLint counterpart ❗
  Check ./tslint-to-eslint-config.log for details.

⚡ 5 new packages are required for this ESLint configuration. ⚡
  npm install @typescript-eslint/eslint-plugin @typescript-eslint/parser eslint eslint-config-prettier eslint-plugin-prefer-arrow --save-dev
❌ 1 error running tslint-to-eslint: ❌
  Error: EISDIR: illegal operation on a directory, read
```

#### Expected Behavior

Shouldn't crash.

#### Reproduction

```shell
mkdir cool
touch cool/index.ts
```

```jsonc
// tsconfig.json
{
  "include": ["cool"]
}
```

As reported by @hyperupcall in #1717.

hyperupcall:
## PR Checklist

-   [x] Addresses an existing issue: fixes #1720

## Overview

If a `tsconfig.json` contains something like the following:

```ts
{
    "include": ["cool"]
}
```

Then, an error shows up, for example, when using `--comments`:

```text
✨ 143 rules replaced with their ESLint equivalents. ✨

⚡ 1 new package is required for this ESLint configuration. ⚡
  npm install eslint-config-prettier --save-dev
❌ 1 error running tslint-to-eslint: ❌
  Error: EISDIR: illegal operation on a directory, read
```

According to the [TypeScript docs](https://www.typescriptlang.org/tsconfig#include), a directory entry for `include` is valid:

> If the last path segment in a pattern does not contain a file extension or wildcard charater, then it is treated as a directory, and files with supported extensions inside that directory are included (e.g. .ts, .tsx, and .d.ts by default, with .js and .jsx if [allowJs](https://www.typescriptlang.org/tsconfig#allowJs) is set to true).

This fixes that case.

### DIFF
--- a/src/adapters/fsFileSystem.ts
+++ b/src/adapters/fsFileSystem.ts
@@ -11,6 +11,14 @@ export const fsFileSystem: FileSystem = {
             return false;
         }
     },
+    directoryExists: async (filePath: string) => {
+        try {
+            const stat = await fs.promises.stat(filePath);
+            return stat.isDirectory();
+        } catch (error) {
+            return false;
+        }
+    },
     readFile: async (filePath: string) => {
         try {
             return (await fs.promises.readFile(filePath)).toString();
--- a/src/comments/collectCommentFileNames.ts
+++ b/src/comments/collectCommentFileNames.ts
@@ -1,3 +1,4 @@
+import { FileSystem } from "../adapters/fileSystem";
 import { SansDependencies } from "../binding";
 import {
     findTypeScriptConfiguration,
@@ -7,6 +8,7 @@ import { uniqueFromSources } from "../utils";
 
 export type CollectCommentFileNamesDependencies = {
     findTypeScriptConfiguration: SansDependencies<typeof findTypeScriptConfiguration>;
+    fileSystem: Pick<FileSystem, "directoryExists">;
 };
 
 export type CommentFileNames = {
@@ -26,12 +28,27 @@ export const collectCommentFileNames = async (
             );
         }
 
+        const includeList = uniqueFromSources(
+            typescriptConfiguration.files,
+            typescriptConfiguration.include,
+        );
+
+        // Remove directories specified in the include list. Ignore
+        // Errors as they'll be handled when reading the file.
+        const includeListFiles = (
+            await Promise.all(includeList.map(dependencies.fileSystem.directoryExists))
+        )
+            .map((isDirectory, i) => {
+                if (isDirectory) {
+                    return null;
+                }
+                return includeList[i];
+            })
+            .filter((item): item is string => typeof item === "string");
+
         return {
             exclude: typescriptConfiguration.exclude,
-            include: uniqueFromSources(
-                typescriptConfiguration.files,
-                typescriptConfiguration.include,
-            ),
+            include: includeListFiles,
         };
     }
 
--- a/src/adapters/fileSystem.stub.ts
+++ b/src/adapters/fileSystem.stub.ts
@@ -1,8 +1,9 @@
 import { FileSystem } from "../adapters/fileSystem";
 import { fn } from "../fn";
 
-export const createStubFileSystem = ({ data = {}, exists = true } = {}) => ({
+export const createStubFileSystem = ({ data = {}, exists = true, isDir = false } = {}) => ({
     fileExists: async () => exists,
+    directoryExists: async () => exists && isDir,
     readFile: async () => data,
     writeFile: fn<FileSystem["writeFile"]>(),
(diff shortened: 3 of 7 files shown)


src/converters/comments/convertFileComments.ts
```diff
@@ -16,6 +16,10 @@ export const convertFileComments = async (
 ) => {
     const fileContent = await dependencies.fileSystem.readFile(filePath);
     if (fileContent instanceof Error) {
+        if (fileContent.code === "EISDIR") {
+            return undefined;
+        }
```
JoshuaKGoldberg, line 21: [Structure] I actually think the approach here isn't what I'd prefer. `collectCommentFileNames` is taking in a `typescriptConfiguration` that looks like:

```json
{
  "compilerOptions": { "target": "es3" },
  "files": ["./cool/index.ts"],
  "include": ["cool"]
}
```

...but with an output `commentFileNames` like:

```js
{
  exclude: undefined,
  include: [ './cool/index.ts', 'cool' ]
}
```

(note that the original TSConfig that made that is just `{ "include": ["cool"] }`. 

I think the right approach to fix this earlier in the system would be to have **`collectCommentFileNames`** filter out directories from its output. Or maybe just not use the `"include"` at all, if `files` has the full list of files?

What do you think?

JoshuaKGoldberg (changes requested):
This works well (yay! thanks!) but I think I'd rather have the issue fixed closer to its source. What do you think?

I'm not against merging as-is if you think what's currently in the PR is a better approach, especially since this repo is old and relatively inactive.

src/converters/comments/convertFileComments.ts
hyperupcall: That sounds good - I changed things so `collectCommentFileNames` handled the filtering of directories

src/adapters/fsFileSystem.ts
```diff
@@ -11,6 +11,14 @@ export const fsFileSystem: FileSystem = {
             return false;
         }
     },
+    directoryExists: async (filePath: string) => {
+        try {
+            const stat = await fs.promises.stat(filePath);
+            return stat.isDirectory();
+        } catch (error) {
```
JoshuaKGoldberg, line 18: Nit:

```suggestion
        } catch {
```

src/comments/collectCommentFileNames.ts
```diff
@@ -26,12 +28,27 @@ export const collectCommentFileNames = async (
...
+        const includeListFiles = (
+            await Promise.all(includeList.map(dependencies.fileSystem.directoryExists))
+        )
+            .map((isDirectory, i) => {
+                if (isDirectory) {
+                    return null;
+                }
+                return includeList[i];
+            })
+            .filter((item): item is string => typeof item === "string");
```
JoshuaKGoldberg, line 47: Nit: there's a util for this already:

```suggestion
            .filter(isTruthy);
```

src/comments/collectCommentFileNames.ts
```diff
@@ -26,12 +28,27 @@ export const collectCommentFileNames = async (
...
+        // Errors as they'll be handled when reading the file.
+        const includeListFiles = (
+            await Promise.all(includeList.map(dependencies.fileSystem.directoryExists))
+        )
+            .map((isDirectory, i) => {
+                if (isDirectory) {
+                    return null;
+                }
+                return includeList[i];
+            })
```
JoshuaKGoldberg, line 46: Nit: could one-line this if you really wanted 😄 

```suggestion
            .map((isDirectory, i) => !isDirectory && includeList[i])
```

JoshuaKGoldberg (approved):
Swell, thanks! This is looking great - and I appreciate you refactoring the internals. 🙌 

I'll apply a couple of nits that honestly don't really change much.

JoshuaKGoldberg:
Oh, I don't have permissions to modify your fork. No worries - it's just nits.

JoshuaKGoldberg:
Published in `tslint-to-eslint-config@2.14.1`. Thanks @hyperupcall! 🥳

PR #1719, opened 2023-08-03
fix(prefer-arrow-functions): Allow `allowStandaloneDeclarations` by default

### LINKED ISSUE #838: prefer-arrow/prefer-arrow-function: prefer arrow to allow top level named functions as default
### 🚀 Feature Request

Currently migration will just output notice that top level named functions are disallowed by the plugin, but default tslint behaviour is to allow them. I think default migration should enable allowStandaloneDeclarations configuration parameter of the prefer-arrow/prefer-arrow-functions rule to match tslint rule behaviour. 

#### Existing Behavior

top level function declarations allowed as default

#### Change Proposal

```json
  "prefer-arrow/prefer-arrow-functions": [
          "error",
          {
            "allowStandaloneDeclarations": true
          }
        ]
``` 
as default migration

hyperupcall:
## PR Checklist

-   [x] Addresses an existing issue: fixes #838

## Overview

Change the default of `prefer-arrow/prefer-arrow-functions` so `allowStandaloneDeclarations` is enabled.

### DIFF
--- a/src/converters/lintConfigs/rules/ruleConverters/only-arrow-functions.ts
+++ b/src/converters/lintConfigs/rules/ruleConverters/only-arrow-functions.ts
@@ -3,13 +3,9 @@ import { RuleConverter } from "../ruleConverter";
 export const convertOnlyArrowFunctions: RuleConverter = (tslintRule) => {
     const notices: string[] = [];
 
-    if (tslintRule.ruleArguments.includes("allow-declarations")) {
-        notices.push("ESLint does not support allowing standalone function declarations.");
-    }
-
     if (tslintRule.ruleArguments.includes("allow-named-functions")) {
         notices.push(
-            "ESLint does not support allowing named functions defined with the function keyword.",
+            "ESLint (eslint-plugin-prefer-arrow plugin) does not support allowing named functions defined with the function keyword.",
         );
     }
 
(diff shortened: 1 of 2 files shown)


src/converters/lintConfigs/rules/ruleConverters/only-arrow-functions.ts
```diff
@@ -17,6 +17,11 @@ export const convertOnlyArrowFunctions: RuleConverter = (tslintRule) => {
         rules: [
             {
                 ...(notices.length !== 0 && { notices }),
+                ruleArguments: [
+                    {
+                        allowStandaloneDeclarations: true,
```
JoshuaKGoldberg, line 22: I poked through the issue (what a throwback!) and https://github.com/TristonJ/eslint-plugin-prefer-arrow/issues/16 as you'd linked - thanks for the info. It look to me like `allowStandaloneDeclarations` should only be output if `"allow-declarations"` is in the original rule. And we can get rid of the _`"ESLint does not support allowing standalone function declarations."`_ notice.

JoshuaKGoldberg (changes requested):
Sorry for marking the issue as accepted without resolving the discussion - looks like the option is opt-in? But thanks for getting started!

hyperupcall:
You're right - the property _is_ opt-in - that's my mistake. The new logic fixes the conversion result

JoshuaKGoldberg (approved):
🆒 thanks!

